PR #22, opened 2020-09-18
feat: page timeout

IonCaza:
### Description

Puts Navigation Timeout in a variable in config.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -12,12 +12,11 @@ const notifications = {
 	email: email.username && email.password
 };
 
-const navigationTimeout = 60000;
-
 const page = {
 	height: 1920,
 	userAgent: 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36',
-	width: 1080
+	width: 1080,
+	navigationTimeout: 60000
 };
 
 const stores = process.env.STORES ?? 'nvidia';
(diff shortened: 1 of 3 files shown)


src/config.ts
```diff
@@ -12,6 +12,8 @@ const notifications = {
 	email: email.username && email.password
 };
 
+const navigationTimeout = 60000;
```
jef, line 15: Let's put this inside the `page` variable

src/config.ts
```diff
@@ -25,5 +27,6 @@ export const Config = {
 	notifications,
 	page,
 	rateLimitTimeout: 5000,
-	stores
+	stores,
+	navigationTimeout
```
jef, line 31: Won't need this anymore if we put inside `page`.

```suggestion

```

jef (changes requested):

src/config.ts
```diff
@@ -15,7 +15,8 @@ const notifications = {
 const page = {
 	height: 1920,
 	userAgent: 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36',
-	width: 1080
+	width: 1080,
+	navigationTimeout: 60000
```
jef, line 19: Oh! Let's also make this an environment variable and add to `README`.

src/config.ts
IonCaza: done

jef (changes requested):

PR #24, opened 2020-09-18
feat: add email test, fix memory leak

DakkJaniels:

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,9 @@ import {Logger} from './logger';
 /**
  * Send test email
  */
-sendNotification('test');
+if (emailTest == 1){
+	sendNotification('test');
+}
 
 /**
  * Starts the bot.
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -5,6 +5,11 @@ import open from 'open';
 import sendNotification from './notification';
 import {Logger} from './logger';
 
+/**
+ * Send test email
+ */
+sendNotification('test');
+
```
jef, line 12: I think we should wrap this behind some sort of environment variable or argument. Just so we don't spam.

Take a look at `config.ts` and see if you want to make an environment variable!

jef (changes requested):

jef:
Thanks for the memory fix! Lets work on the `sendNotification` and go from there.

src/index.ts
DakkJaniels: added an emailTest flag to .env and the config file. Added a test if flag is true to push the notification out. Flag defaults to 0.

PR #26, opened 2020-09-18
feat: add Amazon links

geman220:

PR #28, opened 2020-09-18
Update README.md

geman220:
Adding Amazon to chart once amazon.ts is merged


jef:
Let's append this to #26 (https://github.com/geman220/nvidia-snatcher/blob/patch-3/README.md) and close this out.

PR #29, opened 2020-09-18
feat: update to check if FE in-stock via Digital River

### LINKED ISSUE #2: Use DigitalRiver directly to check for stock on Nvidia
I'm working in Ruby, but here's the basic approach...

Check digital river at a random interval:
> https://api.digitalriver.com/v1/shoppers/me/products/5438481700/inventory-status?apiKey=9485fa7b159e42edb08a83bde0d83dia

Cards should be in stock when `status` is not `PRODUCT_INVENTORY_OUT_OF_STOCK`.

When that happens, open this link to automatically add to cart:
> https://store.nvidia.com/store/nvidia/en_US/buy/productID.5438481700/clearCart.yes/nextPage.QuickBuyCartPage

I expect one of you to sell me a card at price if you get multiple before me 😂 

**NOTE:** `5438481700` is the 3080 US digital river id. I'll post a separate issue with a way of using api.nvidia.com for finding digital river id and opening add to cart when avaliable. You'll want to use this other approach for 3090 launch.

fuckingrobot:
### Description

Fixes #2 
This should reduce load on Nvidia servers, and (hopefully) be faster for y'all.

### Testing

Endpoints tested on my own Ruby script. I can't currently run chromium in my test environment so I'm providing the change for y'all without having tested it within `nvidia-snatcher`.

### DIFF
--- a/src/store/nvidia.ts
+++ b/src/store/nvidia.ts
@@ -6,14 +6,8 @@ export const Nvidia: Store = {
 		{
 			brand: 'nvidia',
 			model: 'founders edition',
-			url: 'https://www.nvidia.com/en-us/geforce/buy/',
-			oosLabels: ['out of stock']
-		},
-		{
-			brand: 'nvidia',
-			model: 'founders edition',
-			url: 'https://www.nvidia.com/en-us/shop/geforce/?page=1&limit=9&locale=en-us&search=3080',
-			oosLabels: ['out of stock']
+			url: 'https://api.digitalriver.com/v1/shoppers/me/products/5438481700/inventory-status?apiKey=9485fa7b159e42edb08a83bde0d83dia',
+			oosLabels: ['PRODUCT_INVENTORY_OUT_OF_STOCK']
 		}
 	],
 	name: 'nvidia'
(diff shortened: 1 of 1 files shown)


src/store/nvidia.ts
```diff
@@ -6,14 +6,8 @@ export const Nvidia: Store = {
...
-			url: 'https://www.nvidia.com/en-us/geforce/buy/',
-			oosLabels: ['out of stock']
-		},
-		{
-			brand: 'nvidia',
-			model: 'founders edition',
-			url: 'https://www.nvidia.com/en-us/shop/geforce/?page=1&limit=9&locale=en-us&search=3080',
-			oosLabels: ['out of stock']
+			url: 'https://api.digitalriver.com/v1/shoppers/me/products/5438481700/inventory-status?apiKey=9485fa7b159e42edb08a83bde0d83dia',
+			oosLabels: ['PRODUCT_INVENTORY_OUT_OF_STOCK']
```
jef, line 10:
```suggestion
			oosLabels: ['product_inventory_out_of_stock']
```

Since we are using `.toLowerCase()`, we should use this!

Otherwise, awesome! Thank you!!!

jef (changes requested):

jef:
Have you witnessed any rate limiting?

jef:
Cool. We got a default 5 second cool down between calls anyway, so it' should be fine.

jef (approved):

PR #33, opened 2020-09-18
fix: small error in `isOutOfStock` logic

vp-cshow:
### Description

While looking into adding some of the other bestbuy AIBs, I noticed that the loop in `isOutOfStock` will always return the check that coincides with the last value of `Store::oosLabels`. This does not align with the logic in the documentation:  

```Checks if DOM has any out-of-stock related text.```

This PR addresses this issue.

There are many different ways to do this, if you have a cleaner idea in mind, feel free to suggest one. I come from a mostly Java background so I'm unsure if node has a cleaner way to do this.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -80,13 +80,7 @@ async function lookup(store: Store) {
  */
 function isOutOfStock(domText: string, oosLabels: string[]) {
 	const domTextLowerCase = domText.toLowerCase();
-	for (const oosLabel of oosLabels) {
-		if (domTextLowerCase.includes(oosLabel.toLowerCase())) {
-			return true;
-		}
-	}
-
-	return false;
+	return oosLabels.some(label => domTextLowerCase.includes(label));
 }
 
 /**
--- a/src/index.ts
+++ b/src/index.ts
@@ -80,12 +80,12 @@ async function lookup(store: Store) {
  */
 function isOutOfStock(domText: string, oosLabels: string[]) {
 	const domTextLowerCase = domText.toLowerCase();
-	let result = false;
 	for (const oosLabel of oosLabels) {
-		result = domTextLowerCase.includes(oosLabel.toLowerCase());
+		if (domTextLowerCase.includes(oosLabel.toLowerCase())) {
+			return true;
+		}
 	}
-
-	return result;
+	return false;
 }
 
 /**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jef** [comment]
I too am not a TS dev by trade! Welcome 😃 !

You're absolutely right. I want to cover my ass and say I didn't write this, but I did merge it and reviewed it 😭 

Looks good to me. Just pass the lint and we'll be good to go!

`npm run lint:fix` does wonders here.

**Comment by vp-cshow** [comment]
Don't worry about it, it wouldn't be software development without random bugs to fix. 

I'm assuming I can add the other best buy cards in a separate PR?

**Review by DakkJaniels — COMMENTED** [review]
nevermind - I see what you meant now. This is good!

File: src/index.ts [hunk 1]
```diff
@@ -80,12 +80,13 @@ async function lookup(store: Store) {
...
-	let result = false;
 	for (const oosLabel of oosLabels) {
-		result = domTextLowerCase.includes(oosLabel.toLowerCase());
+		if (domTextLowerCase.includes(oosLabel.toLowerCase())) {
+			return true;
+		}
 	}
 
-	return result;
+	return false;
```
jef, line 89: If you wanna take it one step further, let's just do this:

```typescript
const domTextLowerCase = domText.toLowerCase();
return oosLabels.some(label => domTextLowerCase.includes(label));
```

Looks great otherwise!

jef:
> The original logic is correct.

Let's pretend our dom text is this: `blah blah out of stock blah blah`

And `oosLabels = ['out of stock', 'not available']`

```javascript
// at first result is false
const result = false

// first loop is 'out of stock'
result = `blah blah out of stock blah blah`.includes('out of stock');

// sets result to true

// now second loop comes around and does a check against 'not available'
result = `blah blah out of stock blah blah`.includes('not available');

// result now gets set to false and we return false.
}
```

src/index.ts
DakkJaniels: Yup, I thought it was merging in the results. Probably could still do that (not sure what the return of .includes is), but just breaking the function once you find our first in-stock hit makes more sense anyway!

vp-cshow: Nice to see more languages are adopting more functional styling; makes things like this more elegant.

DakkJaniels:
Yup, I thought it was merging in the results. Probably could still do that (not sure what the return of .includes is), but just breaking the function once you find our first in-stock hit makes more sense anyway!

jef (approved):

PR #40, opened 2020-09-18
feat: sms notification for usa carriers

thegutuproject:
### Testing

I moved the notification method into the failure section, disabled email notification and watched my phone for incoming SMS

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ First, you're going to need to copy the `.env.example` to `.env`. The current op
 | `EMAIL_USERNAME` | Gmail address; e.g. `jensen.robbed.us@gmail.com` |
 | `EMAIL_PASSWORD` | Gmail password; see below if you have MFA |
 | `STORES` | List of [stores](#Supported-stores) you want to be scraped; optional, default: `nvidia` |
-| `PHONE_NUMBER` | 10 digit phone number. ONLY FOR USA. SMS CHARGES APPLY, for example: `123456789` |
+| `PHONE_NUMBER` | 10 digit phone number. ONLY FOR USA. SMS CHARGES APPLY, for example: `1234567890` |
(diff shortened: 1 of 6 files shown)


jef:
Cool work around. I like it!

README.md
```diff
@@ -67,6 +67,9 @@ First, you're going to need to copy the `.env.example` to `.env`. The current op
 | `EMAIL_USERNAME` | Gmail address; e.g. `jensen.robbed.us@gmail.com` |
 | `EMAIL_PASSWORD` | Gmail password; see below if you have MFA |
 | `STORES` | List of [stores](#Supported-stores) you want to be scraped; optional, default: `nvidia` |
+| `PHONE_NUMBER` | 10 digit phone number. ONLY FOR USA. SMS CHARGES APPLY, for example: `123456789` |
```
jef, line 70: > 10 digit phone number

Uses 9 digit phone number as example 🤣 

Seriously note, look great!!!

README.md
```diff
@@ -67,6 +67,9 @@ First, you're going to need to copy the `.env.example` to `.env`. The current op
...
+| `CARRIER` | Carrier you have. Only supports `sprint, tmobile, att, verizon` |
+| `NOTIFICATION_METHODS` | List of methods you'd like to be notified by, default: `email` |
```
jef, line 72: I'm thinking that we get rid of this environment and don't have this configuration. If the user sets the value, then they should assume that it's going to be sent.

jef (changes requested):

README.md
thegutuproject: 🤣  good catch! and thank you :)

README.md
thegutuproject: I agree, i've removed it in this branch, should be ok now

jef (approved):

PR #44, opened 2020-09-19
fix: sms carrier config, add google carrier

admon84:
### Description

* Fixing .env.example to use `PHONE_CARRIER` so it matches what config.ts is expecting
* Adding `google` carrier for Google Fi users (like myself)

### Testing

I ran the script and verified I can receive SMS on my phone with this code now.


jef (approved):

jef:
This is great! Thank you very much for contributing!

PR #48, opened 2020-09-19
feat(logging): add timestamp

wuruochong:
### Description

Added time to end of each line. Helps identify if program hangs.

### Testing

Ran program with both out of stock and fake in stock messages. Verified that time is at the end of each line.


vp-cshow:
A quick google suggests you can just change the Logger itself instead. It shouldn't be the responsibility of `lookup` to keep track of the time in my opinion.

https://stackoverflow.com/questions/10271373/how-can-i-add-timestamp-to-logs-using-node-js-library-winston

wuruochong:
> A quick google suggests you can just change the Logger itself instead. It shouldn't be the responsibility of `lookup` to keep track of the time in my opinion.
> 
> https://stackoverflow.com/questions/10271373/how-can-i-add-timestamp-to-logs-using-node-js-library-winston

Good point, moved it to logger.

src/logger.ts
```diff
@@ -4,8 +4,10 @@ const prettyJson = format.printf(info => {
 	if (typeof info.message === 'object') {
 		info.message = JSON.stringify(info.message, null, 4);
 	}
-
-	return `${info.level} :: ${info.message}`;
+	
+	const dateObject = new Date();
+	const timeOf = dateObject.toLocaleTimeString();
+	return `${info.level} :: ${info.message} @${timeOf}`;
```
serg06, line 10: I tested this and it looks pretty bad, I think you should change it to

```
return `[${timeOf}] ${info.level} :: ${info.message}`;
```

wuruochong: Thanks for the suggestion! I've updated it.

admon84:
@wuruochong I tried making a similar change with Logger native `format.timestamp()` but I couldn't get passed a lint error with the template literal:  https://github.com/jef/nvidia-snatcher/pull/51/files

Either way, I like this format having the timestamp consistently on the front of the line, like you have it in your latest commit.

jef:
> @wuruochong I tried making a similar change with Logger native `format.timestamp()` but I couldn't get passed a lint error with the template literal: [#51 (files)](https://github.com/jef/nvidia-snatcher/pull/51/files)
> 
> Either way, I like this format having the timestamp consistently on the front of the line, like you have it in your latest commit.

I too was doing this originally, but decided just to leave it out. I think at this point we definitely need timestamps.

jef (approved):
Thanks for your contribution. Looks great 🚀 😄

PR #50, opened 2020-09-19
Adding Zotac for store, Nvidia cartUrl temp fix, Readme update, Newegg/Amazon Captcha workaround, .env.example commented

geman220:
### Description
- Adding Zotac store configurable by zotac in .env
- Adding cartUrl for Nvidia so that API success opens a URL to buy instead of the API result URL
- Updating Readme to show Zotac support
- Added Captcha workaround for Amazon/Newegg looks for text 'Enter the characters you see below' and 'Are you a human?' respectfully to stops false positives
- Commented out .env.example to make the process a bit easier.  Users can rename .env.example to .env without impacting anything.  Add explainations for each .env variable and user can comment out what they do or do not want to use.  For example, several users don't use Slack and the API attempt to connect are throwing up errors.  Disabling by default would remove the extra process and errors.


karthshen (approved):

.env.example
```diff
@@ -1,10 +1,29 @@
...
-NOTIFICATION_TEST="false"
-PAGE_TIMEOUT="30000"
-RATE_LIMIT_TIMEOUT="5000"
-SLACK_CHANNEL="SlackChannelName"
-SLACK_TOKEN="slack-token"
-STORES="bestbuy,bandh,nvidia"
-PHONE_NUMBER="1234567890"
-PHONE_CARRIER="tmobile"
\ No newline at end of file
+// All configuration variables are OPTIONAL.  Uncomment any variables you wish to use or use none of them!
```
jef: I think this type of documentation should live inside the README. In my opinion, it's descriptive enough, but if we want extra help documentation, let's put it inside the wiki.

src/store/nvidia.ts
```diff
@@ -1,7 +1,7 @@
 import {Store} from './store';
 
 export const Nvidia: Store = {
-	cartUrl: '',
+	cartUrl: 'https://www.nvidia.com/en-us/shop/geforce/?page=1&limit=9&locale=en-us&gpu=RTX%203080&gpu_filter=RTX%203090~1,RTX%203080~7,RTX%203070~1,TITAN%20RTX~1,RTX%202080%20Ti~2,RTX%202080%20SUPER~16,RTX%202080~5,RTX%202070%20SUPER~26,RTX%202070~15,RTX%202060%20SUPER~7,RTX%202060~36,GTX%201660%20Ti~29,GTX%201660%20SUPER~4,GTX%201660~11,GTX%201650%20Ti~6,GTX%201650%20SUPER~8,GTX%201650~34',
```
jef, line 4: Since this isn't actually a cart, I think we should remove it.

src/store/zotac.ts
```diff
@@ -0,0 +1,14 @@
+import {Store} from './store';
+
+export const Zotac: Store = {
+	cartUrl: '',
+	links: [
+		{
+			brand: 'zotac',
+			model: 'trinity',
+			url: 'https://store.zotac.com/zotac-gaming-geforce-rtx-3080-trinity-zt-a30800d-10p',
```
jef, line 9: Zotac store is down and has a lot of catchpa right now. I'd suggest updating the `oosLabels`.

jef (changes requested):

.env.example
geman220: Okay, we can remove it.  I just know a lot of people in discord are renaming the .env.example without modifying anything inside of the .env and they are getting lots of errors because they haven't filled in API information etc.

src/store/nvidia.ts
geman220: The potential problem here is that even if Nvidia is detected in stock, it will open the API result on Digital River.  That doesn't really help anyone purchase a card, the previous cartUrl doesn't work any longer so this at least brings you to a page where you can purchase a card.

PR #53, opened 2020-09-19
fix: screenshot size, add screenshot config setting

admon84:
### Description

* Minor tweak to fix screenshot size (1920 x 1080) - Previously width and height were backwards (1080 x 1920)
* Added config option for enabling/disabling screenshots

### Testing

Tested with google.com in the evga config to trigger the screenshot capture for a fake "IN STOCK" hit and verified the screenshot is the correct size.


jef (approved):

jef:
Thanks for the contribution! Looks great :)

PR #54, opened 2020-09-19
fix(store): amazon captcha false-positives

admon84:
### Description

Bandaid fix for amazon captcha page false-positives.

### Testing

Running the script after updating and haven't seen any more false-positives for amazon searches yet.

The amazon captcha page looks like this:
success-1600489110104


admon84:
After running the script for a few hours after this change, I have still seen 2 false positives with amazon captcha pages.  It does appear to happen less frequently, but this is just a temporary bandaid solution anyways.  I won't be offended if this gets closed without merging.

jef (approved):
Haha, good hack. This will be better until we start resolving pages via tags.

Thanks for the input :)

PR #55, opened 2020-09-19
feat(notification): add pushover

therrell:
### Description

Added ability to send push notifications via [Pushover](https://pushover.net/)

Requires user to set up Pushover account for User key and a Pushover Application for Token Key
Free for a week then $5 for lifetime access to push notifications (I found it useful for iOS)

### Testing

N/A

### New dependencies

["pushover-notifications" = "1.2.2"](https://www.npmjs.com/package/pushover-notifications)


src/notification/pushover.ts
```diff
@@ -0,0 +1,23 @@
+/* eslint @typescript-eslint/no-var-requires: "off" */
+import {Config} from '../config';
+import {Logger} from '../logger';
+const Push = require('pushover-notifications');
```
jef, line 4: Does pushover have es6 imports? Maybe @types?

jef:
Otherwise, looks good! Thanks for the contribution :)

src/notification/pushover.ts
therrell: I don't believe so. I couldn't find anything on the [npmjs](https://www.npmjs.com/package/pushover-notifications) page or any examples on github using anything other than `require`. Also tried some workarounds in the code and nothing would work.

I also don't have a ton of experience with all the ES6 standards, so I could be missing something.

jef: No worries. Ya, we can create a type definition (look at `src/types`) and then use the

```typescript
import push = require('pushover-notification')
```

therrell: Thanks, should be fixed now.

jef (approved):
Looks great! Thank you :)

PR #65, opened 2020-09-19
chore(store): add bandh links

gylli251:
Added more bandh cards like i asked on discord.


jef:
Fixed the linting. Thanks for the contribution!

PR #71, opened 2020-09-19
feat(notification): add telegram

### LINKED ISSUE #72: Suggestion: Split out CAPTCHA labels out of `oosLabels`
### Description

Basically split off "out of stock" and "CAPTCHA" labels so that we can log those differently (and see if we just hit a CAPTCHA 100% of the time vs only occassionally)
Would be fairly easy to implement (and I'd gladly PR it myself if it is desired)

### Possible solution

Well, the solution is kind of in the description in this one

Doridian:
### Description

Closes #72

Add Telegram as a notification channel.

Also adds a `.editorconfig` which makes most editors use the correct line endings and indentations

### Testing

Set a valid TELEGRAM_* variables in .env, set NOTIFICATION_TEST=true, saw notification on Telegram

### New dependencies

messaging-api-telegram nodejs package added to package.json


src/store/lookup.ts
```diff
@@ -4,7 +4,15 @@ import {Logger} from '../logger';
 import open from 'open';
 import {Store} from './model';
 import {sendNotification} from '../notification';
-import {isOutOfStock} from './out-of-stock';
+import {isOutOfStock as includesLabels} from './out-of-stock';
+
+function shouldCheckBrand(brand: string) {
```
jef, line 9: Maybe we should have this be `filterBrand`? Little more descriptive maybe.

README.md
```diff
@@ -72,11 +72,13 @@ First, you're going to need to copy the `.env.example` to `.env`. The current op
...
 | `RATE_LIMIT_TIMEOUT` | Rate limit timeout for each full store cycle; optional, default: `5000` |
 | `SLACK_CHANNEL` | Slack channel for posting (e.g., `update`); optional |
-| `SLACK_TOKEN` | Slack API token; optional
+| `SLACK_TOKEN` | Slack API token; optional |
 | `STORES` | [Supported stores](#supported-stores) you want to be scraped; optional, default: `nvidia` |
 | `OPEN_BROWSER` | Toggle for whether or not the browser should open when item is found, default: `true` |
 | `PLAY_SOUND` | Play this sound notification if a card is found.; optional |
 | `SCREENSHOT` | Capture screenshot of page on successful hit; optional, default `true` |
+| `TELEGRAM_BOT_TOKEN` | Telegram bot token; optional |
+| `TELEGRAM_CHAT_ID` | Telegram chat ID; optional |
```
jef, line 81: Perhaps add the `SHOW_ONLY_BRANDS` description?

README.md
```diff
@@ -72,11 +72,13 @@ First, you're going to need to copy the `.env.example` to `.env`. The current op
 | `PHONE_CARRIER` | [Supported carriers](#supported-carriers) for SMS; optional, email configuration required |
 | `RATE_LIMIT_TIMEOUT` | Rate limit timeout for each full store cycle; optional, default: `5000` |
 | `SLACK_CHANNEL` | Slack channel for posting (e.g., `update`); optional |
-| `SLACK_TOKEN` | Slack API token; optional
+| `SLACK_TOKEN` | Slack API token; optional |
 | `STORES` | [Supported stores](#supported-stores) you want to be scraped; optional, default: `nvidia` |
 | `OPEN_BROWSER` | Toggle for whether or not the browser should open when item is found, default: `true` |
 | `PLAY_SOUND` | Play this sound notification if a card is found.; optional |
 | `SCREENSHOT` | Capture screenshot of page on successful hit; optional, default `true` |
+| `TELEGRAM_BOT_TOKEN` | Telegram bot token; optional |
```
jef, line 80: What do you think about changing this to `TELEGRAM_TOKEN` or `TELEGRAM_ACCESS_TOKEN` just to be more related to Telegram?

src/store/lookup.ts
```diff
@@ -32,16 +44,18 @@ export async function lookup(store: Store) {
...
 
 		const bodyHandle = await page.$('body');
 		const textContent = await page.evaluate(body => body.textContent, bodyHandle);
 
 		Logger.debug(textContent);
 
-		if (isOutOfStock(textContent, link.oosLabels)) {
+		if (includesLabels(textContent, link.oosLabels)) {
 			Logger.info(`✖ [${store.name}] ${graphicsCard} is still out of stock`);
+		} else if (link.captchaLabels && includesLabels(textContent, link.captchaLabels)) {
```
jef, line 57: Nice, I like it!

src/store/model/amazon.ts
```diff
@@ -7,49 +7,57 @@ export const Amazon: Store = {
 			brand: 'pny',
 			model: 'xlr8',
 			url: 'https://www.amazon.com/PNY-GeForce-Gaming-Epic-X-Graphics/dp/B08HBR7QBM?ref_=ast_sto_dp',
-			oosLabels: ['currently unavailable', 'enter the characters you see below']
+			oosLabels: ['currently unavailable'],
+			captchaLabels: ['enter the characters you see below']
```
jef, line 11: Good idea. Love it!

jef (changes requested):
Just a few comments. Let me know what you think. I like the `.editorconfig` and the catchpa labels!

README.md
Doridian: It always used to be called bot token. That terminology change must've been fairly recent, so let me fix the env var, sure.

README.md
Doridian: Whoops, forgot that one.

Doridian:
Addressed comments.

README.md
jef: Oh, I actually don't know the terminology at all. I just saw the property name of the function you were calling was `accessToken`, so I figured it was the same.

Whatever you think makes sense then.

jef (approved):

PR #73, opened 2020-09-19
Bugfix - Timeouts caused store lookups to restart from beginning

kylekelso:
### Description

In lookup.ts whenever a page timed out, the original code would exit out of the for loop and restart all looks up from the beginning. First time making a pull request of any kind, so just let me know if anything needs fixing or w/e.

### Testing

Threw an error in the try block of code, and it correctly continues to the next item to lookup now.


jef:
I think it's actually fine the way it is. I think the solution you're looking for is changing `return` to `continue`.

Do you mind changing that?

kylekelso:
Forgot about the continue statement. Should be all set now?

src/store/lookup.ts
```diff
@@ -31,8 +31,7 @@ export async function lookup(store: Store) {
 			await page.goto(link.url, {waitUntil: 'networkidle0'});
 		} catch {
 			Logger.error(`✖ [${store.name}] ${graphicsCard} skipping; timed out`);
-			await browser.close();
```
jef, line 34: I think we'll still need this since it's launching. in the beginning of each loop.

That being said. I do have another PR that will fix this altogether, but let's put this back for now :)

Thanks for all the work!!!

jef (changes requested):

jef:
Someone got you to it :(

Closing in favor of #75

kylekelso:
Oh no worries. Just came back from grabbing dinner. 

Also, I learned a little bit about contributing to git in the process so a win in my book!

jef:
> Also, I learned a little bit about contributing to git in the process so a win in my book!

Awesome to hear. And thanks for helping and bringing it up :)

PR #74, opened 2020-09-19
chore(logging): reformat output

vmnt:
### Description

It's nice to have aligned text. Putting the variable width card name at the end means up til that point the lines will line up. Except different store names (different PR)

### Testing

Ran app, output is as expected
`[12:49:57 PM] info :: ✖ [nvidia] still out of stock: nvidia founders edition`


jef:
> It's nice to have aligned text.

We should look into padding these logs, as I agree it would be nice. The problem here is that the `store.name` is going to vary and will still be misaligned.

If you'd like to look into padding the logger, feel free! Otherwise, we can close and I can put padding the logs on the todo list.

src/store/lookup.ts
```diff
@@ -41,7 +41,7 @@ export async function lookup(store: Store) {
 		Logger.debug(textContent);
 
 		if (isOutOfStock(textContent, link.oosLabels)) {
-			Logger.info(`✖ [${store.name}] ${graphicsCard} is still out of stock`);
+			Logger.info(`✖ [${store.name}] still out of stock: ${graphicsCard}`);
```
jef, line 44: > It's nice to have aligned text.

We should look into padding these logs, as I agree it would be nice. The problem here is that the `store.name` is going to vary and will still be misaligned.

If you'd like to look into padding the logger, feel free! Otherwise, we can close and I can put padding the logs on the todo list.

Thanks for the contribution!!!

jef (changes requested):

src/store/lookup.ts
vmnt: I was going to pad store name (though as you say a general log padder would be even better) in another PR but we should still do this particular change as this is how we would want this particular log line formatted ("OOS: X" instead of "X OOS") regardless of padding. Or do you have in mind that would be resolved by padding each field of the logs in general (maybe I don't know what you mean)

jef: I agree. I like this verbiage better. I think we should put a PR up for padding.

PR #87, opened 2020-09-20
refactor: move `cartUrl` from `Store` to `Link`

geman220:
### Description

As-is the bot only has the ability to have 1 `cartUrl` per site.  The problem is, if a site has multiple links (as almost all do now) you can only assign a single `cartUrl`.  With this change, individual `cartUrl` can be assigned to each link.  By moving the `cartUrl` into the `Links:` we can assign individual `cartUrl` addresses when a product is detected.  I found the add to cart URLs for Best Buy and used it as a proof of concept. 

### Testing

Tested by changing 1 of the Best Buy links to https://www.google.com the bot detected available stock (as expected) but opened the Best Buy add to cart URL instead of the product page.  I also reverted the Best Buy product URL back to the proper URL to ensure it detected no stock as expected.

### New dependencies

None

### DIFF
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -70,7 +70,7 @@ export async function lookup(browser: puppeteer.Browser, store: Store) {
 				await page.screenshot({path: `success-${Date.now()}.png`});
 			}
 
-			const givenUrl = store.cartUrl ? store.cartUrl : link.url;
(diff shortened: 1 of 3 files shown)


jef (approved):

jef:
Makes sense to me!

Thanks!

PR #92, opened 2020-09-20
Amazon stock add

geman220:
### Description

Adding Asus TUF non-oc from US Amazon to bot


src/store/model/amazon.ts
```diff
@@ -51,6 +51,13 @@ export const Amazon: Store = {
 			oosLabels: ['currently unavailable'],
 			captchaLabels: ['enter the characters you see below']
 		},
+		{
+			brand: 'asus',
+			model: 'tuf non-oc',
```
jef, line 56: Looks like this is an OC card.

jef (changes requested):

src/store/model/amazon.ts
geman220: I copied the same link twice because I'm really skilled haha.  Fixed link for non-oc.

Issue #93, opened 2020-09-20
rtx 3080 additions added to amazon.com and newegg


craxyman:
### Description

added remaining rtx 3080 cards not listed on current newegg.ts and amazon.ts

### Possible solution

https://pastebin.com/sNJrwHa6
https://pastebin.com/C3PYYnpz

jef:
Feel free to make a PR! I appreciate the contribution.

PR #95, opened 2020-09-20
feat added office depot store

aitaDev:
### Description
added store to list and created ts store file, also added more cards for newegg from issue #93 

### Testing
this ran for half hour only looping store alone with no false positives. changed out of stock to other text which triggered positive

### DIFF
--- a/src/store/model/index.ts
+++ b/src/store/model/index.ts
@@ -4,6 +4,7 @@ import {Evga} from './evga';
 import {NewEgg} from './newegg';
 import {Amazon} from './amazon';
 import {MicroCenter} from './microcenter';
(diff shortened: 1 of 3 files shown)


jef:
Unfortunately, Office Depot was added here https://github.com/jef/nvidia-snatcher/pull/157. NewEgg links look good though. If you resolve the conflicts, we can get them in.

Thanks!

jef (changes requested):
Unfortunately, Office Depot was added here https://github.com/jef/nvidia-snatcher/pull/157. NewEgg links look good though. If you resolve the conflicts, we can get them in.

Thanks!

PR #96, opened 2020-09-20
chore(store): normalize model names

Wulfre:
### Description

Normalized model names including those in #93, because each contributer wrote the model in their own style.


jef (approved):
Makes sense to me. Thanks!

PR #98, opened 2020-09-20
fix: check response for rate limiting (#58)

### LINKED ISSUE #58: bug: Very high false positive rate due to API key rate limiting
### Description

I just got the script and tried it with this `.env`:

```
NOTIFICATION_TEST="false"
PAGE_TIMEOUT="30000"
RATE_LIMIT_TIMEOUT="5000"
STORES="nvidia"
```

Every 30 seconds (about every 5 requests) I get a false positive because the page returns `Rate limit exceeded for API_KEY [9485fa7b159e42edb08a83bde0d83dia]`. (Confirmed this by looking at the screenshot it took.)

### Possible solution

Ignore the page when it says "Rate limit exceeded". Maybe print a warning to console too.

lukszi:
### Description

Fixes #58 
The issue has been adressed previously, but not been fixed in it's entirety.
This PR fixes the issue by checking for a 429 HTTP response.

### Testing

I ran the script for about half an hour with only the default configuration enabled.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,6 +11,7 @@
     "esModuleInterop": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
-    "forceConsistentCasingInFileNames": true
+    "forceConsistentCasingInFileNames": true,
+	"sourceMap": true
   }
 }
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "esModuleInterop": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
-    "forceConsistentCasingInFileNames": true
+    "forceConsistentCasingInFileNames": true,
+	"sourceMap": true
```
jef, line 15: Nice -- been meaning to do that for some debugging.

jef (approved):
Looks good! Thanks for contributing!!!

PR #108, opened 2020-09-20
feat: update for complex add to cart, fix nvidia

### LINKED ISSUE #38: Nvidia Store add to Cart URL may no longer work
https://github.com/jef/nvidia-snatcher/blob/aa856c712107fadb34630aefd6578dc8950097c7/src/store/nvidia.ts#L4

about an hour ago the add to cart URL would give you the message could not add 3080 to cart as its out of stock... 

If you try the URL now takes you out of the cart and to a new page saying "Sorry, There was a Problem with your request."

Not sure this will still work once cards come in stock

fuckingrobot:
### Description

Fixes #38 once completed. Opening a PR now for some quick feedback before I invest more time into this.

Ideally, it should be possible to take multistep, complex actions _on a per-store basis_ once a product is in-stock. This adds `openCartAction` to the `Link` interface to support these complex actions when needed (it skips the cartUrl/url).

I extracted the DR ID and locale out into constants because they seem to be the only things that will change for foreign stores.

### Testing

~Tested by swapping ID out with 2060 SUPER ID (5379432500) because it's in stock.~ Set up for testing with `debug` and 2060 SUPER.

2060 Super Test


Hari-Nagarajan:
Hey, it looks like you're calling the stock api and then calling the add to cart api if its in stock. You can actually just spam the add to cart api until it gives you a success and then return the carted url. See https://github.com/Hari-Nagarajan/nvidia-bot/blob/master/stores/nvidia.py#L159-L184

mr-sour:
I kinda doubt it. going though digital rivers api they dont have anything that could be used to detect that kind api abuse as is. So it would be on NVIDIA to not only implement rate limiting but also flag suspicious behavior. Do we really expect that kind of feature to be rolled out when theres much lower hanging fruit? I would bet there focusing on just making sure it doesn't crash. I would bet a 3080 order on them doing a bunch of simple bolt on after the fact stuff if they're doing anything at all. Advanced API abuse detection with some kind of rate limiting proxy that flags suspicious calls isn't the kind of thing you can whip up overnight at a company like Nvidia.  If people are ordering just 1 card with how much noise there is on the site its gonna go unnoticed. IMO

jef (approved):

PR #110, opened 2020-09-20
feat(scraping): change lookup impl, add randomize sleep

Doridian:
### Description

This changes the async queue into multiple indepentend timeouts.
It also adds a randomized sleep which might help avoid newegg's bot detection even better (it still CAPTCHA's me at times, so we probably need more on that front).
It also doesn't suffer from the (theoretical) bug of the async queue draining

### Testing

Ran with my usual configuration and observed it correctly creating timeouts


src/config.ts
```diff
@@ -6,7 +6,8 @@ config({path: resolve(__dirname, '../.env')});
 const browser = {
 	isHeadless: process.env.HEADLESS ? process.env.HEADLESS === 'true' : true,
 	open: process.env.OPEN_BROWSER === 'true',
-	rateLimitTimeout: process.env.RATE_LIMIT_TIMEOUT ? Number(process.env.RATE_LIMIT_TIMEOUT) : 5000
+	minSleep: Number(process.env.PAGE_SLEEP_MIN ?? 5000),
```
jef, line 9: :brain:

src/index.ts
```diff
@@ -23,32 +40,10 @@ async function main() {
...
-				Logger.debug(`↗ scraping initialized - ${store.name}`);
-				await lookup(browser, store);
-			} catch (error) {
-				// Ignoring errors; more than likely due to rate limits
-				Logger.error(error);
-			} finally {
-				cb();
-				q.push(store);
-			}
-		}, Config.browser.rateLimitTimeout);
```
jef, line 38: I would keep this queue logic and use the function `getSleepTime` you created.

Is there a reason why we would delete this queue logic?

jef (changes requested):

src/index.ts
Doridian: The queue can theoretically drain even with multiple stores, which would then close the browser.
I don't see the reason for that queue in the first place. It doesn't increase parallelism at all.
And it can never drain (because all functions loop infinitely).

jef (approved):

PR #117, opened 2020-09-20
refactor: extract lookup

jef:
### Description

- Extract lookup process from #110
- Add `await` back to function calls that are `async`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,30 +2,12 @@ import puppeteer from 'puppeteer-extra';
 import stealthPlugin from 'puppeteer-extra-plugin-stealth';
 import adblockerPlugin from 'puppeteer-extra-plugin-adblocker';
 import {Config} from './config';
-import {Store, Stores} from './store/model';
+import {Stores} from './store/model';
 import {Logger} from './logger';
-import {lookup} from './store';
-import {Browser} from 'puppeteer';
+import {getSleepTime, tryLookupAndLoop} from './store';
 
 puppeteer.use(stealthPlugin());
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -41,7 +23,7 @@ async function main() {
 
 	for (const store of Stores) {
 		Logger.debug(store.links);
-		setTimeout(tryLookupAndLoop, getSleepTime(), browser, store);
+		setTimeout(await tryLookupAndLoop, getSleepTime(), browser, store);
```
Doridian, line 26: `await` cannot be used here, at the very least it won't work.

src/store/lookup.ts
```diff
@@ -81,3 +81,20 @@ export async function lookup(browser: Browser, store: Store) {
...
+	Logger.debug(`[${store.name}] Starting lookup...`);
+	try {
+		await lookup(browser, store);
+	} catch (error) {
+		Logger.error(error);
+	}
+
+	const sleepTime = getSleepTime();
+	Logger.debug(`[${store.name}] Lookup done, next one in ${sleepTime} ms`);
+	setTimeout(await tryLookupAndLoop, sleepTime, browser, store);
```
Doridian, line 99: See previous comment

src/index.ts
jef: I thought the same, which I thought was weird. It doesn't complain either way. I wonder if `tsc` throws it out...

jef: Oh, the linter catches it though!

Doridian (approved):

PR #119, opened 2020-09-20
feat: add delay on captcha to try and evade faster

Doridian:
### Description

Adds a delay whenever we get a CAPTCHA. Seems to work well on newegg.
It also moves the sleep functions to a util.ts file


jef (approved):
Nice. I like it. Thanks!!!

PR #120, opened 2020-09-20
feat: card series filter, fix: newegg `oosLabels`

admon84:
### Description

* Adding filterable card series (debug, 3080, and eventually 3090)
* Added 'debug' cards for validating configuration works for "in stock" cards (don't want to miss a chance because the config was wrong - looking at you newegg config)
* Moved oosLabels and captchaLabels to be store props instead of in each link (since they are always the same)

### Testing

* Tested 'debug' and '3080' series on all stores
* Ran lint script

### New dependencies

N/A


admon84:
Newegg oosLabel for `'out of stock'` was throwing "out of stock" message even for in-stock cards.  When I noticed that I figured the "debug" series could be used for testing in-stock cards for each store and make sure it says "IN STOCK" when it should.

src/config.ts
```diff
@@ -59,6 +59,7 @@ const page = {
 };
 
 const store = {
+	showOnlySeries: process.env.SHOW_ONLY_SERIES ? process.env.SHOW_ONLY_SERIES.split(',') : ['3080'],
```
Doridian, line 62: I'd default to `[]` here, same as for brands to defaul to "show all" (or maybe `['3070','3080','3090']` to hide debug cards)?

Doridian (dismissed):
Approved with comment.

src/config.ts
admon84: Yeah, let me change to this `['3070','3080','3090']` as the default.  Using `[]` will include the debug cards and probably not a great idea if someone messes up their environment config

Doridian:
Also please fix lint errors. Looks good otherwise.

Doridian (dismissed):

src/store/includes-labels.ts
```diff
@@ -1,10 +1,16 @@
...
+ * Checks if DOM has any related text.
  *
  * @param domText Complete DOM of website.
- * @param oosLabels Out-of-stock labels.
+ * @param searchLabels Search labels for a match.
  */
-export function includesLabels(domText: string, oosLabels: string[]): boolean {
+export function includesLabels(domText: string, searchLabels: string[]): boolean {
 	const domTextLowerCase = domText.toLowerCase();
-	return oosLabels.some(label => domTextLowerCase.includes(label));
```
jef, line 9: I think this was fine the way it was written. Any reason for the change?

jef (changes requested):
Other than the one comment, I really like what you did here!

Thank you!

src/store/includes-labels.ts
admon84: Agreed, it was fine the way it was.  It was a remainder of another update that got away from me.

admon84: I reverted it back to using `array.some()`. Thanks for the feedback

jef: Ahh, no problem. Just curious if we goofed something up. No biggie. Thanks!

jef (approved):

Ellementz:
By model would be super nice FTW3 anyone?

PR #136, opened 2020-09-20
docs: update pushover variables

### LINKED ISSUE #128: Easy fix: PUSHOVER_USERNAME should be PUSHOVER_USER in readme
### Description

Just change your readme so that PUSHOVER_USERNAME is PUSHOVER_USER as it is referenced in the code.

### Possible solution

See above. 

Easiest fix ever, right?

JoeSchofield:
Fix typo.  PUSHOVER_USERNAME should be PUSHOVER_USER

### Description

Fixes #128 

### Testing

Just verify other code uses USER instead of USERNAME

That code is here: https://github.com/jef/nvidia-snatcher/blob/main/src/config.ts#L40


jef (approved):
Thank you!

PR #137, opened 2020-09-20
feat: twilio phone call integration

Esg450:
### Description
Add twilio phone call integration for sending a reminder phone call when a card becomes available. Good for people who have their phone on vibrate and don't check notifications.

Esg450#3153

### New dependencies
"@types/twilio": "^2.11.0"
"twilio": "^3.49.2"


src/notification/twilio.ts
```diff
@@ -0,0 +1,25 @@
...
+import twilio from 'twilio';
+
+const client = twilio(Config.notifications.twilio.sid, Config.notifications.twilio.auth);
+
+export function requestTwilioCall() {
+	let toNumber = Config.notifications.phone.number;
+
+	if (!toNumber.startsWith('+1')) {
+		toNumber = `+1${toNumber}`;
+	}
```
jef, line 12: Instead of having any of this logic, the user should be responsible for providing a correct country prefix, as well as their number. I do understand we are piggybacking off another config, but we should probably just update the docs to note this if it's required by Twilio.

jef (changes requested):

src/notification/twilio.ts
Esg450: Good call on that. I do think that if we piggyback off the PHONE_NUMBER variable, there will be and problems with sms if we request the users to put the country code before the number to use Twilio since it's using email to send the text. I think there's two ways we can go with this:

1. Create a new config variable for the country code and add it to the PHONE_NUMBER programmatically (how it's working now, but with the new country code var).

2. Convert the sms functionality to use Twilio instead. That will get rid of any carrier checking, but the downside is that it will require the creation of a Twilio account. 

Let me know which route you'd prefer and I can implement either one.

jef: Definitely the former if anything. I'd be curious to do some testing with the email including the country code.

Esg450: Just did some testing with sprint and verizon and sending the text with the +1 before the number was a no go. I'll go ahead and push the changes with the new variable shortly.

jef: Thanks for testing!

lukechen526:
@Esg450 thank you for working on this - do you think you could add Twilio SMS as an option/ alternative to phone call?

Esg450:
@lukechen526 I'll leave that up to @jef on how we would want to implement that. It's easy enough to add into the code, but I'm not sure how we should handle the configuration for that as to not overcrowd the config or confuse users. Let me know on discord if theres any ideas for that.

jef:
Do you mind resolving these conflicts?

As for the SMS, what other configuration would you need? Should the token work the same for the phone call?

Esg450:
Just merged and made the changes. I spoke to @lukechen526 on discord and went ahead and added the TWILIO_MODE config that will accept the `sms` and/or `phone` modes to give an alternative to email texting.

Esg450:
Just merged with all the changes again @jef , let me know if there are any other requested changes.

PR #139, opened 2020-09-20
fix: memory leak due to adblocker

Doridian:
### Description

Plugs a memory leak due to some architectural oversights in the adblocker plugin.
It must be detached manually from every page, or it will keep a stale reference alive and prevent it from being GC'd

### Testing

Heapdumps and checking via `htop` for memory allocations

### DIFF
--- a/src/adblocker.ts
+++ b/src/adblocker.ts
@@ -1,0 +1,11 @@
+import {Page} from 'puppeteer';
+import {PuppeteerExtraPluginAdblocker} from 'puppeteer-extra-plugin-adblocker';
+
+export const adBlocker = new PuppeteerExtraPluginAdblocker({
+	blockTrackers: true
+});
+
+export async function disableBlockerInPage(page: Page) {
+	const blockerObject = await adBlocker.getBlocker();
+	await blockerObject.disableBlockingInPage(page);
+}
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -5,7 +5,7 @@ import open from 'open';
 import {Store} from './model';
 import {sendNotification} from '../notification';
 import {includesLabels} from './includes-labels';
-import {delay, getSleepTime} from '../util';
+import {closePage, delay, getSleepTime} from '../util';
 
 /**
  * Returns true if the brand should be checked for stock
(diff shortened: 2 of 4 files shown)


src/adblocker.ts
```diff
@@ -0,0 +1,12 @@
...
+import {PuppeteerExtraPluginAdblocker} from 'puppeteer-extra-plugin-adblocker';
+
+export const adBlocker = new PuppeteerExtraPluginAdblocker({
+	blockTrackers: true
+});
+
+export async function closePage(page: Page) {
+	const blockerObject = await adBlocker.getBlocker();
+	await blockerObject.disableBlockingInPage(page);
+	await page.close();
```
gcssabbagh, line 11: The `page.close()` call shouldn't be in here. Should keep it in lookup.ts. Only adblocker related actions should reside here.

src/adblocker.ts
```diff
@@ -0,0 +1,12 @@
+import {Page} from 'puppeteer';
+import {PuppeteerExtraPluginAdblocker} from 'puppeteer-extra-plugin-adblocker';
+
+export const adBlocker = new PuppeteerExtraPluginAdblocker({
+	blockTrackers: true
+});
+
+export async function closePage(page: Page) {
```
gcssabbagh, line 8:
```suggestion
export async function disableBlocker(page: Page) {
```

src/store/lookup.ts
```diff
@@ -62,7 +63,7 @@ async function lookup(browser: Browser, store: Store) {
 			response = await page.goto(link.url, {waitUntil: 'networkidle0'});
 		} catch {
 			Logger.error(`✖ [${store.name}] ${graphicsCard} skipping; timed out`);
-			await page.close();
+			await closePage(page);
```
gcssabbagh, line 66:
```suggestion
			await disableBlocker(page);
			await page.close()
```

src/store/lookup.ts
```diff
@@ -96,7 +97,7 @@ async function lookup(browser: Browser, store: Store) {
 			sendNotification(givenUrl, link);
 		}
 
-		await page.close();
+		await closePage(page);
```
gcssabbagh, line 100:
```suggestion
		await disableBlocker(page);
		await page.close()
```

src/adblocker.ts
Doridian: Ideally? Yes, sure. However I'd rather have a single function to tear down a page so we don't duplicate two calls allover the place.
Maybe move this out into a helper function in `lookup.ts`?

gcssabbagh: Yeah that would work too

Doridian:
Moved to util function in `util.ts` and kept adblocker file limited to adblocker

jef (approved):

jef:
Awesome. Great find!

PR #141, opened 2020-09-20
feat: invert logic

admon84:
### Description

* Inverting logic to check for 'add to cart' labels within an visible element container scope, instead of checking for out-of-stock  and captcha labels within the body text

### Testing

* Tested script with all stores and seeing expected results for in-stock debug cards and out-of-stock 3080s

### New dependencies

* N/A


src/store/lookup.ts
```diff
@@ -66,34 +66,49 @@ async function lookup(browser: Browser, store: Store) {
...
+		if (store.labels.captcha) {
+			const captchaHandle = await page.$(store.labels.captcha.container);
+			const captchaContent = await page.evaluate(element => element.textContent, captchaHandle);
 
-			if (Config.browser.open) {
-				await open(givenUrl);
+			if (includesLabels(captchaContent, store.labels.captcha.labels)) {
+				Logger.warn(`✖ [${store.name}] CAPTCHA from: ${graphicsCard}. Waiting for a bit with this store...`);
+				await delay(getSleepTime());
+				await page.close();
```
Doridian, line 103: Use `closePage` utility function

src/store/lookup.ts
```diff
@@ -66,34 +66,49 @@ async function lookup(browser: Browser, store: Store) {
...
+				if (Config.browser.open) {
+					await open(givenUrl);
+				}
+
+				sendNotification(givenUrl, link);
 
-			if (Config.page.capture) {
-				Logger.debug('ℹ saving screenshot');
-				await page.screenshot({path: `success-${Date.now()}.png`});
+				await page.close();
```
Doridian, line 91: Use `closePage` utility function

src/store/lookup.ts
admon84: Done

src/store/lookup.ts
admon84: Done

Doridian (approved):
The logic seems sound. The ifs look a bit odd, but I can't quite think of how to better implement this.
Need to fix merge conflicts, tho.

admon84:
Agreed on the ifs looking odd.  I changed the lookup method to use early-exit style ifs, each section will closePage and continue the loop.  Do you think this way is better?

Doridian:
Yeah I like it a bit more now, at least it is uniform. But I think maybe we should move the contents of the loop into its own function such that

Pseudocode-y bit here:
```
const page = ...;
try {
    await lookupCard(store, page, ....); // This contains what the loop used to contain
} catch (e) {
    Logger.error(e);
}
await closePage(page);
```

This allows us to just `return` from the sub-function and not have to repeat the closePage thing multiple times (and it also prevents any edge-case where the browser context might not be terminated correctly due to whatever error

admon84:
I'm working on updating the code to use sub-functions for lookup logic

jef:
Moved as a draft PR. Feel free to **Ready for review** when you're done!

Thanks for the work ❤️

admon84:
I think this is ready for review again.  The logic seems to be working for all cases when I'm testing, debug cards all showing in stock and 3080's are showing out of stock currently.  I've done my best to factor in new store additions (officedepot, newegg-ca, asus).  

The biggest problem childs for this logic are microcenter and bestbuy, who's sites aren't as easy to scrape for 'add to cart' labels as others because of how they handle their buttons.

Doridian (approved):

jef:
> The biggest problem childs for this logic are microcenter and bestbuy, who's sites aren't as easy to scrape for 'add to cart' labels as others because of how they handle their buttons.

Should we keep the old logic then for this type of lookup?

jef:
image

I also don't see a lot of the containers you're mentioning in the text. Not sure if it's me or what :shrug:

admon84:
Try checking the debug page for cards in stock. All of the containers should be found on those pages for in-stock products, which should match what the 3080 pages will contain when they go in-stock with add to cart buttons:

Screenshot from 2020-09-22 07-55-20

admon84:
> Should we keep the old logic then for this type of lookup?

I did get them working in the end, but they took the most time to figure out is all because they were more non-conventional.

Screenshot from 2020-09-22 07-59-06

Screenshot from 2020-09-22 08-00-20

One thing worth calling out is how Microcenter logic works, its checking the page for a list of "(in stock)" labels, and since the cards can only be bought in stores, it might trigger "IN STOCK" if like Duluth GA has a card in stock, but when you open the page it might now show "In Stock" unless you have your store preference set to the right store.  I think this is the right way to go in the end though

admon84:
I did a lot more testing and can say the results are looking good to me

Doridian (approved):

jef (approved):

jef:
Big change! I appreciate the contribution on this one :smile:

PR #148, opened 2020-09-21
chore: add types for `brand` and `series`

scttcper:
### Description

Closes https://github.com/jef/nvidia-snatcher/issues/30

Adds types for Link type brands and series. I wanted to filter by brand and they weren't listed anywhere. Figured i'd add the type while i'm in there. Opted not to use enums because the code itself is fine how it is.


scttcper:
model has some messy names like `gaming x trio - duplicate` that i'm not sure i want to handle right now

jef:
> model has some messy names like `gaming x trio - duplicate` that i'm not sure i want to handle right now

Model definitely needs some normalizing. Perhaps a list like you have. We can normalize those in a later PR.

I really like this idea though. Thanks for your contribution!

src/store/model/store.ts
```diff
@@ -1,8 +1,8 @@
 import {Browser} from 'puppeteer';
 
 export interface Link {
-	series: string;
-	brand: string;
+	brand: 'TEST' | 'asus' | 'evga' | 'gigabyte' | 'pny' | 'msi' | 'nvidia' | 'zotac';
+	series: 'debug' | '3070' | '3080' | '3090';
```
jef, line 5: I think we should settle on a brand/series for debug/test. I honestly don't know if either have any functional implications, but my guess no. Do you mind changing to this?

```suggestion
	brand: 'test:brand' | 'asus' | 'evga' | 'gigabyte' | 'pny' | 'msi' | 'nvidia' | 'zotac';
	series: 'test:series' | '3070' | '3080' | '3090';
```

Also, was there any more elaboration on:

> Opted not to use enums because the code itself is fine how it is.

If you wanted to keep this PR to a minimal, that makes sense. Just seeing if there was any technical reason you say as well.

Anyway, thanks for the contribution! I appreciate it!

jef (changes requested):

src/store/model/store.ts
scttcper: Yeah just keeping the PR minimal

admon84 (dismissed):

jef (approved):

PR #154, opened 2020-09-21
feat: Add German Nvidia store

### LINKED ISSUE #84: Crawl German Nvidia.de Store Page
I have the product id for the german nvidia store page, is it possible to crawl that instead of the USA site?

ColinFarL:
### Description

Fixes #84 

Add the Nvidia.de store, using the in-and-ru-store API to check stock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ Here is a list of variables that you can use to customize your newly copied `.en
 | Micro Center | `microcenter`|
 | Newegg | `newegg`|
 | Nvidia | `nvidia`|
+| Nvidia.de | `nvidia-de`|
 
 #### Supported carriers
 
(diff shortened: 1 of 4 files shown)


DarkeleEXE:
If i try this version I always get a success, cause of the following errors: 504 gateway time-out, 503 service temporarily unavailable and another error which i cant reproduce at the moment

ColinFarL:
Interesting. This API is what Nvidia's de-de site actually uses for its own stock check, so it's as reliable as the site itself.
(Notably, the calls from the Nvidia store page are also timing out, so it's currently unable to check its own stock on any items.)

Either way, I'll add failure states for 500 status codes when I get home.

DarkeleEXE:
I just got another error instead of an API result, but the API seems way more stable now. 5 minutes ago I had only failures:


Exception(s) were found for route(s): 
	0: Response code 429 mapped as failure.

ColinFarL:
> Seems like its not working right, I've modified the .ts with your latest .ts

Have you pulled from main? The file it's complaining about was just added a couple of hours ago.

DarkeleEXE:
For me this new version is working 100%, zero false positives since running it

ColinFarL:
While this _is_ (finally) working, a better implementation is happening in #162, so I'll close this PR.

PR #162, opened 2020-09-21
feat: set country in config, login to nvidia when starting

fuckingrobot:
### Description

Fixes #161. Refactors Nvidia to hide more of the logic. Adds support for a bunch of countries to Nvidia (with add to cart!) and support for 3090 (despite no DR IDSs).

### Testing

Tested on `usa` and `belgium`. Use `.env` as `README` shows.

image

PR #163, opened 2020-09-21
Added GB nvidia store (extendable method)

HMellor:
### Description
Added a locale based implementation for checking different nvidia store locations. Only GB and US implemented but easily extendable if you know the DigitalRiver ID for your region.

### Testing
Checked that generated `digitalRiverStockUrl`s were valid


HMellor:
Just noticed this is a duplicate of #162, I'll close this

PR #178, opened 2020-09-21
chore(store): evga oos label

designgears:
Updated labels to avoid false-positive when their site goes down.

image


designgears:
ah, already a pr for it

PR #183, opened 2020-09-21
feat(api): add rudimentary web control panel

Doridian:
### Description

Adds a web panel. Need to set `WEB_PORT` for it to bind. This has no authentication, so don't run this on the public internet without a firewall


geman220:
I want it, nice work!

admon84 (approved):
Great idea!

jef:
I'm gonna pull this down today and check it out. Thanks Doridian!

jef:
@Doridian it's awesome. I love it. Good for the start of something awesome!

Now I think for an initial release of this, we just need a few changes. I would say that changing anything in the web, should overwrite what is in `.env` (store, brand, series wise). As of now, changing something in the web UI just appends what's already there. Perhaps intended, but maybe should clear out anything we haven't selected.

Not as necessary, but could be possible with minimal effort?: I'd also suggest making the list not a multi-selectable list, but rather checkboxes. Perhaps more user friendly.

---

I think this is _almost_ there and this will be an awesome addition.

Doridian:
I agree that checkboxes are easier to use, actually. Good idea!

As for updating the `.env`, I need to find a good way of doing that. Since essentially I need to map the huge JS object back to the variables that created it. Maybe at first just support what the UI supports. (As you might have seen in the code, the "API" supports updating the entire config, not just the fields I surface in the UI).
So at first maybe just update `STORES` and `SHOW_ONLY_*` type variables?

jef:
> So at first maybe just update STORES and SHOW_ONLY_* type variables?

Agreed, I think that's plenty for now.

Also to clarify (which may not be needed); You may not need to manipulate the `.env` file, just overwrite the maps that read in the `.env` initial.

Can't wait to see it! Thanks for doing this @Doridian <3

jef:
You ready for me now @Doridian?

Doridian:
@jef Any time really. Didn't yet find a way I like for it to persist its config, tho. With the `.env` config style.

jef:
Okay sounds good. I'll pull it down when I get home from vacation and give it a run. From there, we can get this baby merged in!

jef:
@Doridian I think I'm gonna pull this in today. I'm going to change the base branch to develop and start working on top of it. Probably release as 2.0.0 and will come along with some name changes I have in mind.

jef:
Crap... I really want this in here. I'll pull it down in the near future and get it merged in. This is my top priority.

Thank you @Doridian for the patience on this one.

jef:
Okay... Last time, I promise. I won't bring in anything else. Let's resolve these conflicts and iterate on this one. Sorry @Doridian 😞

jef (approved):
It's going in!!!!

PR #186, opened 2020-09-22
feat(notification): add pushover priority setting

LBRapid:
### Description

Allow the priority of pushover messages to be set. This is useful because it can allow users to have notifications come in as critical alerts on iOS for example.

### Testing

Tested with `npm run test:notification`


Vendigroth:
Adding a sound parameter that defaults to 'pushover' can be useful too.

jef:
I like it! Sorry for all the merge conflicts. If you can get those resolved, we can get this in!

Thank you very much for the contribution.

LBRapid:
@jef I believe this should be good to go now

jef (approved):

PR #193, opened 2020-09-22
refactor: use index appropriately

jef:
### Description

Moves quite a bit of files in favor of exporting from an `index.ts`.


Doridian:
Honestly, within a project, I'm not sure if I like the usage of `index.ts` files. Usually a pattern I only observe in libraries.
I was actually thinking a PR of removing all `index.ts` files that only consist of `export * from './file';`

jef:
> Honestly, within a project, I'm not sure if I like this usage of `index.ts` files. Usually a pattern I only observe in libraries.
> I was actually thinking a PR of removing all `index.ts` files that only consist of `export * from './file';`

Not a bad idea. I'm not a TS expert by any means. @andrewmackrodt, have any thoughts?

jef:
Closing for now, but will keep for discussion.

PR #209, opened 2020-09-22
feat: add chromium sandbox skipping

andrewmackrodt:
This PR adds an optional environment variable **TRUSTED** which can be used to skip puppeteer/chromium's sandbox usage, it defaults to **false**. Relates to https://github.com/jef/nvidia-snatcher/issues/114 but our approach is slightly different, i.e. this change does not force the insecure setting unless set by the user.

Edit: linking to sample Docker usage using an image pushed to the DockerHub registry: https://github.com/andrewmackrodt/dockerfiles/tree/master/nvidia-snatcher#docker


fuckingrobot (dismissed):

jef (approved):

jef:
I really like this! Although, I think `TRUSTED` could be used in a lot of different contexts. Do you mind if we prefix this environment variable with `BROWSER_` or something similar?

Thanks!

andrewmackrodt:
> Do you mind if we prefix this environment variable with `BROWSER_` or something similar?

Sure, I've added the `BROWSER_` prefix. The variable on the config object has stayed as `isTrusted` as I feel `Config.browser.isTrusted` is verbose enough.

jef:
> `Config.browser.isTrusted` is verbose enough.

Agreed! Sorry, I wasn't clear there, but it all looks good from here!

Thank you!!

src/config.ts
```diff
@@ -5,6 +5,7 @@ config({path: path.resolve(__dirname, '../.env')});
 
 const browser = {
 	isHeadless: process.env.HEADLESS ? process.env.HEADLESS === 'true' : true,
+	isTrusted: process.env.TRUSTED ? process.env.BROWSER_TRUSTED === 'true' : false,
```
jef, line 8:
```suggestion
	isTrusted: process.env.BROWSER_TRUSTED ? process.env.BROWSER_TRUSTED === 'true' : false,
```

jef (approved):
Looks great! Thanks!

I don't necessarily want to start managing a Dockerfile quite yet because it'll open another can of worms, so this is a great start for those interested.

Thanks!

PR #220, opened 2020-09-22
feat(store): add evga model

levelfifty:
Added lower tier gaming model of FTW3 and updated descriptions.


jef (approved):
Nice! Thanks :smile:

Issue #116, opened 2020-09-20
Sound player issue


jdkennedy45:
> $ npm start
> 
> nvidia-snatcher@1.4.0 start C:\Users\Jake\Documents\nvidia-snatcher
> npm run build && node build/index.js
> 
> 
>  nvidia-snatcher@1.4.0 build C:\Users\Jake\Documents\nvidia-snatcher
>  rimraf ./build && tsc
> 
> INFO: Could not find files for the given pattern(s).
> INFO: Could not find files for the given pattern(s).
> INFO: Could not find files for the given pattern(s).
> INFO: Could not find files for the given pattern(s).
> INFO: Could not find files for the given pattern(s).
> INFO: Could not find files for the given pattern(s).
> INFO: Could not find files for the given pattern(s).
> INFO: Could not find files for the given pattern(s).
> [2:06:40 PM] info :: ✖ [amazon] still out of stock: pny xlr8
> [2:06:41 PM] info :: ✖ [newegg] still out of stock: asus tuf
> ....

Hi I'm just confused on what this means. It prints it out every time I start the program. But the program seems to operate normally after..?

jef:
Not sure what's going on here. Can you post your `.env`?

m1se:
Maybe its related to the naming of the .env file?

jdkennedy45:
> DISCORD_NOTIFY_GROUP="@here"
> DISCORD_WEB_HOOK="XXXXXXXXXXXXX"
> EMAIL_USERNAME="XXXXXXXXXXXXXXXXXX"
> EMAIL_PASSWORD="XXXXXXXXXXXXXXX"
> HEADLESS="true"
> LOG_LEVEL="info"
> OPEN_BROWSER="true"
> PAGE_TIMEOUT="30000"
> PHONE_NUMBER="XXXXXXXXXXXXXXX"
> PHONE_CARRIER="sprint"
> PLAY_SOUND="C:\Users\Jake\Downloads\swiftly.mp3"
> PUSHOVER_TOKEN="XXXXXXXXXXXXXX"
> PUSHOVER_USER="XXXXXXXXXXXXXXX"
> RATE_LIMIT_TIMEOUT="5000"
> SHOW_ONLY_BRANDS="evga,zotac,aorus,asus,msi,gigabyte,pny"
> SLACK_CHANNEL="SlackChannelName"
> SLACK_TOKEN="slack-token"
[...]

PR #233, opened 2020-09-23
chore(misc): tiny improvements

serg06:
Preface: Please let me know if I need a better title.

### Description

Improvement 1: I alphabetically sorted some lists in `README.MD`.

Improvement 2: I added better handling for when the user enters an invalid store name. If they enter an invalid name, it'll print a warning instead of throwing an exception, and if they don't enter any valid names, the app will quit right away.

Improvement 3: To stop more issues like #116 popping up, I added some descriptive logging around the big block of "INFO: Could not find files for the given pattern(s)." output.

Discord: sergo#8847

### Testing

1. Tried with `STORES="nvidia"`
2. Tried with `STORES="invalid-store"`


geman220:
Looks good to me!

jef (approved):
Great! Thanks for cleaning some of this up. I doubt this will stop people from duplicating issues, but it's definitely worth the extra work here 😆 

Appreciate the work!

PR #234, opened 2020-09-23
fix(store): adorama captcha config

admon84:
### Description

* Restoring captcha label for adorama store config

### Testing

* Ran script against adorama store

### New dependencies

* N/A


jef (approved):
Looks good! Thanks!

PR #248, opened 2020-09-23
chore(notification): additional ca carriers

bubba1337:
bell,fido,rogers,kudo, and virgin mobile

### Description

 Adds additional Canadian phone carriers for SMS service
 I am Canadian

### Testing

 Tested locally for bell and is working using npm run test:notification


jef (approved):
Sure, why not. If they're broken, we can remove them.

Thanks!

PR #249, opened 2020-09-23
feat(stores): add 3090 for bestbuy, newegg

Doridian:


jef (approved):
Let's gooooooooo!!!!!

PR #259, opened 2020-09-24
chore(notification): warn user of improper sms and email config

vp-cshow:
### Description

Noticed that some people in the discord were trying to use the SMS notifications without setting up email as well; quick check just to make sure the user knows.


jef (approved):

jef:
Thanks!

PR #261, opened 2020-09-24
feat: filter models

### LINKED ISSUE #30: Filter card brand and model
### Description

It would be nice to filter by brand and model across all sites.

### Possible solution

Create global keys for brands and models and make a filter function.

admon84:
### Description

* Feature: Filter models (similar to brand/series)
** Logging info on selected filters during startup
* Fixes #30

### Testing

Test 1 - no filtering on models
```
SHOW_ONLY_BRANDS="nvidia,evga,asus,msi"
SHOW_ONLY_MODELS=""
SHOW_ONLY_SERIES="3080"
STORES="nvidia,bestbuy,evga,asus,amazon,newegg"
```
Result:
Screenshot from 2020-09-23 20-07-12

----

Test 2 - filter on models
```
SHOW_ONLY_BRANDS="nvidia,evga,asus,msi"
SHOW_ONLY_MODELS="founders edition,rog strix"
SHOW_ONLY_SERIES="3080"
STORES="nvidia,bestbuy,evga,asus,amazon,newegg"
```
Results:
Screenshot from 2020-09-23 20-11-52

### New dependencies

N/A


vp-cshow:
Suggestion: 

I feel that this feature will be **very** error prone by the end user. Maybe we should use a string similarity algorithm like levenshtein distance (I have a Java implementation hat I can probably dig up...) or another lib to prevent something breaking because of something like the user entering rogstrix instead of rog strix?

admon84:
> Suggestion:
> 
> I feel that this feature will be **very** error prone by the end user. Maybe we should use a string similarity algorithm like levenshtein distance (I have a Java implementation hat I can probably dig up...) or another lib to prevent something breaking because of something like the user entering rogstrix instead of rog strix?

I see what you're saying, but I wouldn't want to over-complicate it either.  With string comparisons it's common to compare lower/upper case user input to a value to eliminate case mismatching.

Maybe in this case (for filtering models) we should remove all whitepsace when comparing the model name in the store config to what the user typed into their .env config

vp-cshow:
Sounds like a plan; just giving my 2c.

admon84:
Latest commit adds whitespace sanitization, so at least it will be a little forgiving with user input having spacing not matching the store config: "gaming x trio" vs "gamingx trio" vs "gamingxtrio" 
Screenshot from 2020-09-23 20-34-00

src/store/filter.ts
```diff
@@ -0,0 +1,62 @@
+import {Config} from '../config';
```
jef, line 1: This was on my todo list to refactor this out. Love it!

jef (approved):
Looks good! I like it. Nicely done with the regex.

One comment I will make is that doing regex over and over again like this isn't the best thing to do.

One thing we should do next is do the filtering all before the program starts to parse through the pages, that way we've done all the setup we need to do. That's outside of this, but just a comment.

Thanks!

PR #265, opened 2020-09-24
feat: add link series to error message for better context

gcssabbagh:
### Description

Couldn't tell which nvidia link was timing out so I added the series to the error message

### Testing

Just ran snatcher again and all those nvidia timeouts now show the series

### New dependencies
N/A


jef (approved):
Yep, makes sense. Surprised I didn't have this in the `Print` variable.

Thanks for doing this!

PR #268, opened 2020-09-24
feat(store): multiple inStock labels config

### LINKED ISSUE #322: Half wrong container class name for the Nvidia store
### Description

There are two urls that you are using to hit Nvidia's store(3080 used as an example):
URL 1: https://www.nvidia.com/en-us/shop/geforce/gpu/?page=1&limit=9&locale=en-us&category=GPU&gpu=RTX%203080
URL 2: https://www.nvidia.com/en-us/geforce/graphics-cards/30-series/rtx-3080

I noticed that by changing the inStock text to 'out of stock' to test an alert, and removing URL 1. It doesn't trip an alert.
This is because the container class name currently used, `.main-container`, doesn't exist on URL 2. 

### Possible solution

The easy fix is to change it to `general-container`. I Found that this trips an alert on both URLs.

### Side Note/Warning
However I do want to point out that when cards were available this past Wednesday, Nvidia made big changes to their site. I had URL 1 refreshing the whole time and i noticed that the listing for the 3080 never showed a button with `Add to Cart`, but rather `Check Availability`. Clicking that redirected you to URL 2. That was just my observation, correct me if i'm wrong.

The only reason i brought this issue up is because I believe URL 2 is more reliable, However Nvidia added an antibot measure on URL 2 where they delay the visibility of the out of stock button by about 10-15 seconds(not sure if you noticed that). No worries, because I tested this and you guys had that all figured out with Puppeteer's `networkidle0`

admon84:
### Description

Closes #322 

* Changed store.labels config to allow multiple inStore containers
    * This helps for configs like `nvidia` store where a product list page and product details page are being searched, but the `'.main-container'` selector only works for the products list page (and a different container selector needs to be used for the product details page `'.js-product-item .btn-cart'`)

### Testing

* I added a second debug card in the `nvidia` store config for testing, and verified both debug pages are matching the 'add to cart' buttons and showing the "IN STOCK" alert

### New dependencies

N/A

### DIFF
--- a/src/__test__/notification-test.ts
+++ b/src/__test__/notification-test.ts
@@ -11,10 +11,10 @@ const link: Link = {
 
 const store: Store = {
 	labels: {
-		inStock: {
-			containers: ['test:container'],
+		inStock: [{
+			container: 'test:container',
 			text: ['test:text']
-		}
+		}]
 	},
 	links: [link],
 	name: 'test:name'
--- a/src/__test__/notification-test.ts
+++ b/src/__test__/notification-test.ts
@@ -12,7 +12,7 @@ const link: Link = {
 const store: Store = {
 	labels: {
 		inStock: {
-			container: 'test:container',
+			containers: ['test:container'],
 			text: ['test:text']
 		}
 	},
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -91,18 +91,15 @@ async function lookupCard(browser: Browser, store: Store, page: Page, link: Link
 }
 
 async function lookupCardInStock(store: Store, page: Page) {
-	const stockHandle = await page.$(store.labels.inStock.container);
-
-	const visible = await page.evaluate(element => element && element.offsetWidth > 0 && element.offsetHeight > 0, stockHandle);
-	if (!visible) {
-		return false;
+	/* eslint-disable no-await-in-loop */
+	for (const container of store.labels.inStock.containers) {
+		if (await lookupCardHasContent(page, container, store.labels.inStock.text)) {
+			return true;
+		}
 	}
+	/* eslint-enable no-await-in-loop */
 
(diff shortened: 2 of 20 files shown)


jef:
Closes #322

jef:
This looks great! If you are able to resolve these conflicts, this is going in next!

Thank you very much @admon84 :)

andrewmackrodt:
#270 added support for array of inStock labels, would the below be enough to cover this use case?

```patch
Index: src/store/model/nvidia.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/store/model/nvidia.ts	(revision 07bd246e876cd27df1b5019af5ee8613bb5368f0)
+++ src/store/model/nvidia.ts	(date 1601131903909)
@@ -6,10 +6,13 @@
 			container: 'body',
 			text: ['are you a human?']
 		},
-		inStock: {
+		inStock: [{
 			container: '.main-container',
 			text: ['add to cart']
-		}
+		}, {
+			container: '.js-product-item .btn-cart',
+			text: ['add to cart']
+		}]
 	},
 	links: [
 		{
@@ -19,6 +22,12 @@
 			url: 'https://www.nvidia.com/en-us/shop/geforce/gpu/'
 		},
 		{
+			brand: 'test:brand',
+			model: 'test:model',
+			series: 'test:series',
+			url: 'https://www.nvidia.com/en-us/geforce/graphics-cards/rtx-2060-super/'
+		},
+		{
 			brand: 'nvidia',
 			model: 'founders edition',
 			series: '3080',

```

jef:
> #270 added support for array of inStock labels, would the below be enough to cover this use case?

I thought this got added already and I must've misunderstood the intent. At this point, the current labels are `Element[]`, thus abstracting out all of these properties.

Makes sense to close this one out then.

admon84:
I agree, this one can be closed out.

PR #270, opened 2020-09-24
feat: enhanced lookup behaviour

andrewmackrodt:
### Description
Am opening up this PR for discussion as it has some useful features. Much of the work was done to support several UK stores and ended up duplicating #141 which wasn't merged at the time. I've since rebased and tried to make all changes fully backwards compatible with the structure prior to #141 and with the current HEAD.

- Fixes runtime errors caused by incorrect devDependencies when running with `npm install --only=production`
- Uses ts-node to speed up dev workflow a little, also allows end user running without `build/` folder
- Adds some missing card brands that are available at some stores (in the UK at least)
- Adds the ability to disable the puppeteer adblocker extension on a per-store basis - this was required for scan.co.uk where every request after the first one would fail otherwise
- Enhances lookup behaviour:
  - Status codes >= 400 fail immediately
  - 403 (Forbidden) responses are retried with exponential backoff to allow IP rate limits to reset
  - Empty responses are sent to the debug log
  - Restores `outOfStock` and `bannedSeller` behaviours (see below)
  - Requires all checks to pass before declaring a card in stock, i.e. if any of `inStock`, `outOfStock`, `bannedSeller` or `captcha` are defined, they must pass
  - Adds support for multiple `Element` lookups, i.e. multiple `container`, `text` definitions
  - _Restores support for "legacy" array of string checks (can be removed - work was started on this before the changes were merged which changed to the Element style)_
  - Adds the ability to determine a stores links at runtime, see eBuyer and Scan examples here https://github.com/andrewmackrodt/nvidia-snatcher/compare/pr-enhancements...develop - note: the current implementation only works for the first page of results, this is fine for the stores given as an example

### Testing
Tested with the following `.env` file:

```env
OPEN_BROWSER="false"
SCREENSHOT="false"
SHOW_ONLY_SERIES="test:series,3080"
STORES="adorama,amazon,amazon-ca,asus,bandh,bestbuy,bestbuy-ca,evga,evga-eu,microcenter,newegg,newegg-ca,officedepot,zotac"
```

To test the behaviour of runtime detected store links please try the `develop` branch of my fork https://github.com/andrewmackrodt/nvidia-snatcher

### New dependencies

| package | description |
|-|-|
| cheerio | DOM parser with js/css like selectors, used by dynamic link feature |
| nodemon | (dev) monitors the project for changes and automatically restarts it |
| ts-node | (dev) compiles typescript on-the-fly |

### Changed dependencies
| package | description |
|-|-|
| @slack/web-api | moved from devDependencies to dependencies |
| discord-webhook-node | moved from devDependencies to dependencies |
| play-sound | moved from devDependencies to dependencies |


jef:
This is really good. I'll have to look at it in the AM so I can have a closer look. Thanks for contributing!!!

nodemon.json
```diff
@@ -0,0 +1,8 @@
+{
+  "exec": "TS_NODE_FILES=true ts-node ./src/index",
```
andrewmackrodt, line 2: Note: this needs to be changed as it doesn't work for Windows users.

package.json
```diff
@@ -7,7 +7,7 @@
     "build": "rimraf ./build && tsc",
     "lint": "xo",
     "lint:fix": "xo --fix",
-    "start": "npm run build && node build/index.js",
+    "start": "nodemon --config nodemon.json",
```
jef, line 10: This will be better I think.

```suggestion
    "start": "npm run build && node build/index.js",
    "start:nodemon": "nodemon --config nodemon.json",
```

src/index.ts
```diff
@@ -37,14 +38,20 @@ async function main() {
...
 			store.setupAction(browser);
 		}
 
+		if (store.linksBuilder) {
+			await fetchLinks(store, browser);
+		}
+
 		setTimeout(tryLookupAndLoop, getSleepTime(), browser, store);
 	}
+	/* eslint-enable no-await-in-loop */
```
jef, line 54: I think you're fine using [this solution](https://eslint.org/docs/rules/no-await-in-loop)

```suggestion
	const promises = [];
	for (const store of Stores) {
		Logger.debug(store.links);
		if (store.setupAction !== undefined) {
			store.setupAction(browser);
		}
		
		if (store.linksBuilder) {
			promises.push(fetchLinks(store, browser));
		}

		setTimeout(tryLookupAndLoop, getSleepTime(), browser, store);
	}
	await Promise.all(promises)
```

src/store/fetch-links.ts
```diff
@@ -0,0 +1,40 @@
...
+			}
+
+			const docElement = cheerio.load(text).root();
+			const links = store.linksBuilder!.builder(docElement, series);
+			Logger.info(Print.message(`FOUND ${links.length} ITEMS`, series, store, true));
+			Logger.debug(JSON.stringify(links, null, 2));
+			store.links = store.links.concat(links);
+		});
+	}
+	/* eslint-enable no-await-in-loop */
```
jef, line 39: Same solution as above. Shouldn't be a problem with wait time outs since this only happens once per store in the beginning of boot up.

jef (commented):
Nice! I like a lot of these changes. All seem great.

Just a few comments. Other than that, this should definitely go in!

src/store/lookup.ts
```diff
@@ -44,7 +60,41 @@ async function lookupCard(browser: Browser, store: Store, page: Page, link: Link
...
+	if (!backoff) {
+		backoff = {count: 0, time: Config.browser.minBackoff};
+		storeBackoff[store.name] = backoff;
+	}
+
+	if (response?.status() === 403) {
+		Logger.warn(Print.backoff(link, store, backoff.time, true));
+		await delay(backoff.time);
+		backoff.count++;
+		backoff.time = Math.min(backoff.time * 2, Config.browser.maxBackoff);
```
jef, line 78: > 403 (Forbidden) responses are retried with exponential backoff to allow IP rate limits to reset

This keeps it from going to infinite then, correct?

andrewmackrodt:
Thanks for the review @jef. I'm thinking of breaking out the "Adds the ability to determine a stores links at runtime" feature to another PR as I've modified beyond the scope of what's discussed in here, primarily it can run multiple times during tryLookupAndLoop https://github.com/andrewmackrodt/nvidia-snatcher/blob/develop/src/store/lookup.ts#L172. The reason for doing so was for yesterdays 3090 stock availability check before the digital-river-product-id was known, and I'm happy to say it worked; as is the feature generally which successfully found 3090 stock at currys and box today.

src/store/lookup.ts
andrewmackrodt: Yes, up to a default of one hour.

package.json
andrewmackrodt: What do you think of "dev" instead of "start:nodemon"?

ts-node could be invoked directly in the start command (eliminating the need to create build/ and without nodemon's process reloading). I think it's a little nicer to not have to create build/ in it's entirety every time the app is started but it would still be creating files in /tmp under the hood so doesn't make much difference I guess.

andrewmackrodt: I've restored the behaviour of "start" and added "dev" as described above, with a short description in README.md. If you strongly prefer "start:nodemon" please let me know and I'll update it.

src/store/fetch-links.ts
andrewmackrodt: Have pushed a change using `Promise.all(...)`.

src/index.ts
andrewmackrodt: Have pushed a change using `Promise.all(...)`.

package.json
jef: Nope, `dev` is fine with me.

package.json
```diff
@@ -22,12 +23,16 @@
   },
   "homepage": "https://github.com/jef/nvidia-snatcher#readme",
   "dependencies": {
+    "@slack/web-api": "^5.12.0",
```
jef, line 26: Weird that we installed this and the others with `--save-dev`. Good catch.

jef (approved):
Alright! Looks good. I'm going to go in after this and change the `npm run dev` to a `start:dev` task. I hope you don't mind or take it personal, because this PR is great and don't want to keep requesting changes from ya.

Thanks!

andrewmackrodt:
`start:dev` is find with me, `start:nodemon` is a little long, is it pronounced `node-mon` or `no-demon`, I never know :sweat_smile:.

serg06:
I feel like bannedSellers needs to be configurable in some way, e.g. an option in .env to enable/disable it. I want to ban all 3rd-party sellers but I feel like some people wouldn't want that.

jef:
@serg06, #383 just got merged. I think you'll want to do that now.

PR #289, opened 2020-09-24
chore: add `.exe` to `.gitignore`

locmai0808:
### Description
Some people may add audio players to the root of repo (like me), so I thought it's worth to add .exe to `.gitignore`


jef (approved):

jef:
Cool. Thanks!

PR #294, opened 2020-09-24
feat: low bandwidth mode

### LINKED ISSUE #290: Add low bandwidth mode
### Description

By blocking images and fonts, the snatcher can load pages a little faster and reduce the overall traffic the script downloads. This will not impact the ability of the snatcher to review the page for updates. 

However, this feature can't be enabled currently as the `puppeteer-extra-plugin-adblocker` package already intercepts all requests in order to block those. A relevant discussion about multiple request interceptions via `puppeteer-extra` can be found here: https://github.com/berstend/puppeteer-extra/issues/91

### Possible solution

Users could be provided flags for the adblocker and/or low bandwidth modes and informed that low bandwidth mode requires the adblocker to be disabled. Blocking images and fonts should improve speed and reduce traffic more than the adblocking, but I'm not certain if adblocking was enabled for other reasons.

joshuahiggins:
### Description

Closes #290 

Enables a low bandwidth mode that blocks image and font requests in order to load pages faster and reduce overall network traffic. Since images and fonts have no impact on JS or HTML, this change should cause zero impact to the capabilities of this script. This will make screenshots slightly uglier on some sites.

Due to a conflict with the adblocker, adblocking is disabled when `LOW_BANDWIDTH` flag is set to true. Relevant information on conflict: https://github.com/berstend/puppeteer-extra/issues/91

### Testing

Ran with headless mode disabled and commented out the page closing so that pages could be inspected and confirmed to work properly. Also reviewed network traffic running both with and without `LOW_BANDWIDTH` enabled. Traffic appears to be reduced in the 30-50% range.

### New dependencies

- `puppeteer-extra-plugin-block-resources@^2.2.7`


geman220:
I like it!  One thing, I know we are trying to keep things alphabetized so you might want to move your `LOW_BANDWIDTH=""` down a couple of lines in `.env-example`.  Other than I'm a fan!  More options is always better and if someone doesn't want to use it they don't have to =P.

joshuahiggins:
Hah! I noticed it wasn't alphabetized, but I still meant to have it 1 line lower so it was sorta alphabetized. Not sure how I missed that.

jef:
> Due to a conflict with the adblocker

Makes sense to me!

I wonder if this should be a default of `true`. Did you find any odd behaviors when in use?

joshuahiggins:
Weirdest issue I found was that with image resources blocked, Amazon still manages to load an extremely low resolution default image for most products. Kinda threw me off. No negative findings though.

jef:
> Weirdest issue I found was that with image resources blocked, Amazon still manages to load an extremely low resolution default image for most products. Kinda threw me off. No negative findings though.

Alright, sounds good. Well... I'll run on my side for the #stock notifications and move it to default if we find that it's pretty solid.

I appreciate the work! Great contribution.

jef (approved):

serg06:
Nice! Can't wait to use it.

Looking at [this list](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/ResourceType), it looks like we can block some other resource types as well?

Ideas:
- `imageset`
- `media`

remusao:
Hey there,

Looking at the incompatibility with the adblocker, I wanted to suggest that maybe the adblocker itself can be used to block fonts and images. This way you would not need to pick one mode or the other.

I can look into it if there is interest.
Best,

jef:
Yeah! Look into that if you'd like. I think both options are viable and probably give similar results.

remusao:
> Yeah! Look into that if you'd like. I think both options are viable and probably give similar results.

Sure this might not be too different. I assume that if both are enabled more requests will be blocked (since the adblocker will also block some scripts and other types).

PR #296, opened 2020-09-25
fix: denver `microcenter` name

cbandrow:
### Description

Doesn't fix any known issues that I found, but for `microcenter.ts`, the map of locations had an incorrect location: 
`['devin', '181'],` should instead be `['denver', '181'],`

The change should be fixed with this PR. Confirmed the "181" store location via the URL for the Denver store: 
image


src/store/model/microcenter.ts
```diff
@@ -10,7 +10,7 @@ const microCenterLocationToId: Map<string, string> = new Map([
 	['chicago', '151'],
 	['columbus', '141'],
 	['dallas', '131'],
-	['devin', '181'],
```
jef, line 13: Lol... No idea. Sorry about that. Shout out to the CU Boulders :wink:

jef (approved):

jef:
Thank you!

src/store/model/microcenter.ts
cbandrow: lol it's all good, thanks for putting this whole thing together. Saw it as an easy way to boost my really limited github score.

PR #306, opened 2020-09-25
fix: fix Polish locale ID in nvidia-api

keton:
### Description

This fixes Polish locale ID. Was pl-pSl, correct value pl-pl.


jef:
Fixed in https://github.com/jef/nvidia-snatcher/pull/314

Thanks for pointing this out! :smile:

PR #314, opened 2020-09-25
fix: `nvidia-api`

### LINKED ISSUE #267: Nvidia Snatcher Not Working for...Nvidia
Nvidia changed their page so that "Out of Stock" is always displayed even when a buy button is available. As a result, I don't think the snatcher knows when there is stock. I missed two drops today (one in the afternoon and one about an hour ago) because no notifications were sent and no alarm went off. The latest apparently was up for 15 minutes, so there's some bug as to why it's missing it. I've been getting notifications for other sites (though mostly false alarms).

andrewmackrodt:
Fixes #267 - uses the new API to check for stock status and the new API add-to-card action which generates the cart redirect URL. I've only tested this with a 2060 Super by changing the `great_britain.fe3080Id` value to that of an in stock 2060 Super and adding to cart and opening browser works. The actual stock detection change has been working for me for over 36 hours, i.e. it successfully has tracked 3080 and 3090 stock in the UK store, I actually purchased a 3080 6am UK time yesterday.

AFAIK the only thing missing compared to the previous behaviour was opening a page when the app is started. I've not found a way to get the cart redirection url without first adding a product to the basket. Also not tested is if previous checkouts are overwritten if someone has a low (or unset) `IN_STOCK_WAIT_TIME` setting.

Test script for currency update:

```ts
import {regionInfos} from './store/model/nvidia-api';
import {get} from 'request';
import {nvidiaStockUrl} from './store/model/helpers/nvidia';
import {promisify} from 'util';

async function verifyNvidiaProductsHavePricing() {
	for (const regionInfo of Array.from(regionInfos)) {
		const [country, data] = regionInfo;
		if (!data.fe3080Id) {
			continue;
		}

		const storeUrl = nvidiaStockUrl(data.fe3080Id, data.drLocale, data.currency);
		let productData: Record<string, any> | undefined;

		try {
			const response = await promisify(get)({
				headers: {
					'user-agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.157 Safari/537.36'
				},
				url: storeUrl
			});

			productData = JSON.parse(response.body);

			if (typeof productData !== 'object' || productData === null) {
				throw new Error('Not a JSON object');
			}
		} catch (e) {
			console.error(`Failed to retrieve product information for country ${country}`);
			console.error(e.message);
			continue;
		}

		const price = productData.products?.product[0]?.pricing?.listPrice;

		if (price) {
			console.info(`Found price information for country ${country} => ${price.currency} ${price.value}`);
		} else {
			console.warn(`Failed to find price information for country ${country}`);
		}
	}
}

void verifyNvidiaProductsHavePricing();

```


src/store/model/helpers/nvidia.ts
```diff
@@ -1,89 +1,73 @@
...
 import open from 'open';
 import {timestampUrlParameter} from '../../timestamp-url-parameter';
 
-const nvidiaApiKey = '9485fa7b159e42edb08a83bde0d83dia';
-
 function getRegionInfo(): NvidiaRegionInfo {
 	const country = Array.from(regionInfos.keys()).includes(Config.store.country) ? Config.store.country : 'usa';
 
-	const defaultRegionInfo: NvidiaRegionInfo = {drLocale: 'en_us', fe2060SuperId: 5379432500, fe3080Id: 5438481700, fe3090Id: null, nvidiaLocale: 'en_us'};
+	const defaultRegionInfo: NvidiaRegionInfo = {currency: 'USD', drLocale: 'en_us', fe2060SuperId: 5379432500, fe3080Id: 5438481700, fe3090Id: null, nvidiaLocale: 'en_us'};
```
thegutuproject, line 12: We can populate the fe3090Id here: 5438481600

src/store/model/helpers/nvidia.ts
```diff
@@ -1,89 +1,73 @@
...
-		timestampUrlParameter();
-}
+async function addToCartAndGetLocationRedirect(page: Page, sessionToken: string, productId: number): Promise<string> {
+	const url = 'https://api-prod.nvidia.com/direct-sales-shop/DR/add-to-cart';
+
+	page.removeAllListeners('request');
+
+	await page.setRequestInterception(true);
+
+	page.on('request', interceptedRequest => {
```
jef, line 41: Nice, this will speed this up quite a bit. Was looking for this type of API in puppeteer. Nice find!

jef (changes requested):
> AFAIK the only thing missing compared to the previous behaviour was opening a page when the app is started

Yep! This is the right approach as of now (see https://github.com/Hari-Nagarajan/nvidia-bot/pull/128).

I really appreciate you putting in this work! This was next on my to-do list and you've got it all sorted!

src/store/model/helpers/nvidia.ts
andrewmackrodt: Sorry if I'm being dense, but what is the requested change?

jef: Oh.. I didn't send :o

src/store/model/helpers/nvidia.ts
jef, line 12: May have been a copy paste error, but I think we do have a 3090 code here. You've got it down below at least! 😄 

```suggestion
	const defaultRegionInfo: NvidiaRegionInfo = {currency: 'USD', drLocale: 'en_us', fe2060SuperId: 5379432500, fe3080Id: 5438481700, fe3090Id: 5438481600, nvidiaLocale: 'en_us'};
```

jef (changes requested):

src/store/model/helpers/nvidia.ts
jef: Lol I had two reviews open... Whoops. Thanks!

src/store/model/helpers/nvidia.ts
andrewmackrodt: It is `null` in the existing `HEAD`. I've modified the function to remove duplication by checking if the regions map contains the country and falling back to usa if not.

src/store/model/helpers/nvidia.ts
```diff
@@ -7,10 +7,17 @@ import open from 'open';
...
-	const country = Array.from(regionInfos.keys()).includes(Config.store.country) ? Config.store.country : 'usa';
+	let country = Config.store.country;
+	if (!regionInfos.has(country)) {
+		country = 'usa';
+	}
+
+	const regionInfo = regionInfos.get(country);
+	if (!regionInfo) {
+		throw new Error(`LogicException could not retrieve region info for ${country}`);
+	}
```
jef, line 18: I think you can remove this as you already check in lines 11-13?

jef: Better approach though. Thanks!

jef (changes requested):

src/store/model/helpers/nvidia.ts
andrewmackrodt: > I think you can remove this as you already check in lines 11-13?

I think it needs to be checked anyway or the TypeScript compiler will correctly throw an error that regionInfo may be undefined, e.g. if usa was renamed or removed.

jef:
> which should the .env contain as store? nvidia or nvidia-api

This would mean that you could use both. Either one. `nvidia` will scrape the site, the `nvidia-api` will call the APIs

andrewmackrodt:
@manybot `STORES=nvidia-api`.

niagi:
Might as well leave the swedish FE3090 product ID here: 5438761600

src/store/model/helpers/nvidia.ts
jef: Based on that assumption, I suppose you could do the opposite then and do one call.

```typescript
	if (!regionInfo) {
		country = 'usa';
	}
```

andrewmackrodt: I may not be following, but the only way this will work without 2 if conditions is if we silence the typescript compiler, e.g.
```ts
function getRegionInfo(): NvidiaRegionInfo {
	let regionInfo = regionInfos.get(Config.store.country);
	if (!regionInfo) {
		regionInfo = regionInfos.get('usa')
	}

	return regionInfo!;
}
```
It has no way to tell that `usa` will always exist in the `regionInfos` map, and it's IMO correct to behave that way.

src/store/model/nvidia-api.ts
```diff
@@ -12,25 +13,25 @@ export interface NvidiaRegionInfo {
...
-	['luxembourg', {drLocale: 'fr_fr', fe2060SuperId: 5394902700, fe3080Id: 5438795700, fe3090Id: null, nvidiaLocale: 'fr_fr'}],
-	['netherlands', {drLocale: 'nl_nl', fe2060SuperId: 5394903500, fe3080Id: 5438796700, fe3090Id: null, nvidiaLocale: 'nl_nl'}],
-	['poland', {drLocale: 'pl_pl', fe2060SuperId: null, fe3080Id: 5438797700, fe3090Id: null, nvidiaLocale: 'pl_pSl'}],
-	['portugal', {drLocale: 'en_gb', fe2060SuperId: null, fe3080Id: 5438794300, fe3090Id: null, nvidiaLocale: 'en_gb'}],
-	['russia', {drLocale: 'ru_ru', fe2060SuperId: null, fe3080Id: null, fe3090Id: null, nvidiaLocale: 'ru_ru'}],
-	['spain', {drLocale: 'es_es', fe2060SuperId: null, fe3080Id: 5438794800, fe3090Id: null, nvidiaLocale: 'es_es'}],
-	['sweden', {drLocale: 'sv_SE', fe2060SuperId: null, fe3080Id: 5438798100, fe3090Id: null, nvidiaLocale: 'sv_se'}],
-	['usa', {drLocale: 'en_us', fe2060SuperId: 5379432500, fe3080Id: 5438481700, fe3090Id: 5438481600, nvidiaLocale: 'en_us'}]
+	['austria', {currency: 'EUR', drLocale: 'de_de', fe2060SuperId: null, fe3080Id: 5440853700, fe3090Id: null, nvidiaLocale: 'de_de'}],
```
jef, line 16: I wanted to remove one of these locales since it's only necessary we use one now, but I will update that in another chore PR.

andrewmackrodt:
@niagi I'd prefer if anything not directly related to changing the api implementation was a different PR. If nvidia change this implementation again and this PR gets reverted for some reason, then additional changes such as 3090 ids would also be lost.

src/store/model/helpers/nvidia.ts
jef: Ah, got it. No problem. I haven't pull your branch, so I guess I haven't seen what the compiler spits out yet. Makes sense to me!

src/store/model/nvidia-api.ts
andrewmackrodt: I'd considered that, but this is a good point. I haven't tested whether adding to cart works for Canada and Poland which have different `nvidiaLocale` to `drLocale`. A shame the nvidia site is broken at the moment or I'd test this.

src/store/model/helpers/nvidia.ts
jef: Ah. I see now. I come from a different language so the idiosyncrasies are still catching up.

I was thinking usually when you call `T.get()` or us `[]`, if it can't find the object, it'll throw an exception. I see that `Map.get()` in TS will just return an `undefined` object, rather than throw.

My assumption is that we would do something like this:

```
function getRegionInfo(): NvidiaRegionInfo {
	const regionInfo = regionInfos.get(Config.store.country);
	if (!regionInfo) {
		return regionInfos.get('usa')
	}

	return regionInfo;
}
```

But I see the `TypeAssertion` error now.

---

Below isn't a request to change, just some feedback would be nice for a learning experience on my end if you didn't mind!

Would something like this run fine?

```typescript
function getRegionInfo(): NvidiaRegionInfo {
	const regionInfo = regionInfos.get(Config.store.country);
	if (!regionInfo) {
		return <NvidiaRegionInfo>regionInfos.get('usa')
	}

	return regionInfo;
}
```

It compiles, but as long as we know `'usa'` is a provided country, we could. This could still be dangerous as `'usa'` may "suddenly" disappear.

I suppose in that case we could create a wrapper function around `Map.get()` for getting a default `NvidiaRegionInfo` and avoid altogether!

src/store/model/nvidia-api.ts
jef: We'll clean it up if ends up being the case. This is solid for now though!

jef (approved):

miralize:
Incredible work with both this & #270 @andrewmackrodt, @jef 👏

src/store/model/helpers/nvidia.ts
andrewmackrodt: That would work, TS allows type coercion so long as the type is legal. Another way to write that would be `regionInfos.get('usa') as NvidiaRegionInfo`, or `regionInfos.get('usa')!` to tell the compiler it's definitely defined. The last one would be most applicable to this example, type coercion is only needed to tell the compiler the result is a one of multiple defined types that are valid.

jef: Got it. So best case is to use the implementation you had, but next best is the `regionInfos.get('usa')!`. Makes sense to me. I suppose it lacks a keyword [`unsafe`](https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html) :wink: 

Thanks for the explanation!

Issue #312, opened 2020-09-25
Only skip checking per IN_STOCK_WAIT_TIME for the item that is in stock


shaver:
### Description

Right now, if a store has any of its possible items in stock, all further checks for that store, for all items, are blocked for IN_STOCK_WAIT_TIME seconds. This means that I have to decide between
- getting a notification every 10 seconds for some scalper's $2500 3080 that I'm not going to touch
- not finding out for a while about any other stock that appears on amazon-ca

### Possible solution

It would be nice if the block applied only to the item that was actually found.

jef:
Fixed in https://github.com/jef/nvidia-snatcher/commit/c7a716f981976a76afe61a4d985bd6fe4343595b and https://github.com/jef/nvidia-snatcher/pull/325

Related: https://github.com/jef/nvidia-snatcher/issues/312

PR #325, opened 2020-09-26
fix: in stock wait time

### LINKED ISSUE #315: IN_STOCK_WAIT_TIME not working as intended
### Description

Set IN_STOCK_WAIT_TIME in .env file to several  different options including 3600 which should make it so the same link isn't notified more than once an hour but was still getting the same link notification once every 2 minutes.

Screenshot_17

jef:
### Description

Fixes #315, #312

Wrongly was checking for `store.name` instead of `link.url` during refactor of `IN_STOCK_WAIT_TIME`.


serg06:
Thank you!

jef:
@serg06 no problem :) thanks for report!

PR #330, opened 2020-09-26
chore: add missing nvidia drid for all series/regions

andrewmackrodt:
Adds all detectable digital-river-id. The lookup of these values was automated using [this script](https://github.com/andrewmackrodt/nvidia-snatcher/blob/develop/src/store/model/helpers/nvidia-dr.ts) in my `develop` branch. You can run it via the command:

```sh
node -r ts-node/register/files src/scripts/detect-all-digital-river-id
````
I've only manually validated a few results but the script checks that the value we have for the 3080 series matches what it automatically detects, this should give us some confidence it is doing the correct region lookup.


jef:
Tested it out and actually works pretty well. I also saw your some of your link building for the other stores. Looks cool there too 😄

Makes sense to pull this in. I'm sure people will report otherwise 😼

jef (approved):

PR #339, opened 2020-09-27
docs: update supported countries

andrewmackrodt:
Adds norway (was previously missing) and updates 3090 and test card statuses. Removes documentation for russia as a supported country as there are no useable product ids and no support generally^.

^ we only use english texts when checking for stock and the russian store uses a different direct-sales-shop endpoint


jef (approved):
💀 RIP 🇷🇺

PR #344, opened 2020-09-27
feat(notification): add Twillio notification provider

### LINKED ISSUE #342: Support sending notifications via Twilio
### Description

Since the built in SMS support only appears to support US carriers it would be nice to also allow sending SMSs via Twilio. This way users in any country can recieve SMS notifications. 

### Possible solution

Integrate the Twilio node client as a new notification provider (PR Incoming).

WillsB3:
### Description
Adds Twilio (https://www.twilio.com/) as a notification provider so that users outside of the US can receive stock notifications via SMS.

Fixes #342
Discord: Wills#6846

### Testing

- [x] Ran `npm run test:notification` and received the test SMS.
- [x] Tested using my own Twilio account and received notifcation for `test:series` in sock via `SHOW_ONLY_SERIES="test:series"`.

### New dependencies

- [Twilio](https://www.npmjs.com/package/twilio)


src/notification/notification.ts
```diff
@@ -53,6 +54,11 @@ export function sendNotification(link: Link, store: Store) {
 		sendTelegramMessage(link, store);
 	}
 
+	if (notifications.twilio.accountSid && notifications.twilio.authToken) {
+		Logger.debug('↗ sending twilio message');
+		sendTwilioMessage(link, store);
+	}
```
WillsB3, line 60: Adding this causes a code complexity linting warning. Not sure on how you'd prefer this be solved. Does this function need a refactor?

jef: Nah, that's ok. We need to separate these out at some point. It's probably just complaining because of the line length within this function.

jef (approved):
Straight forward! Looks good to me.

Thanks!

methane-frost:
this is making the latest build fail, as it appears a Twilio username is now required?

rrbailey89:
can confirm, latest build wont start now.

src/notification/twilio.ts
```diff
@@ -0,0 +1,26 @@
+import {Link, Store} from '../store/model';
+import {Logger, Print} from '../logger';
+import {Config} from '../config';
+import twilio from 'twilio';
+
+const config = Config.notifications.twilio;
+
+const client = twilio(config.accountSid, config.authToken);
```
WillsB3, line 8: @jef Sorry - only just seen this was merged. Didn't get GitHub email notifications for some reason.

I've seen reports of users having issues since this has been merged, saying that they are getting an error that Twilio credentials are needed even when they haven't set any of the Twilio options.

I think this line is the cause. We should move it so we are only initalizing the client inside `sendTwilioMessage`. I'll try and get a PR up later.

andrewmackrodt: @WillsB3 another user provided a fix in #349 which has been merged.

WillsB3: Ah great, thanks.

Sorry for the trouble

PR #346, opened 2020-09-27
fix(notification): discord false triggers

andrewmackrodt:
Fixes output `↗ sending discord message` when no `webHookUrl` defined.

### DIFF
--- a/src/notification/notification.ts
+++ b/src/notification/notification.ts
@@ -38,7 +38,7 @@ export function sendNotification(link: Link, store: Store) {
 		sendDesktopNotification(link, store);
 	}
 
-	if (notifications.discord.webHookUrl) {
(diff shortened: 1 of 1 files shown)


src/notification/notification.ts
```diff
@@ -38,7 +38,7 @@ export function sendNotification(link: Link, store: Store) {
 		sendDesktopNotification(link, store);
 	}
 
-	if (notifications.discord.webHookUrl) {
+	if (notifications.discord.webHookUrl.length > 0) {
```
jef, line 41: Right... Oops. Happened when I changed this to be lists. I'll start just putting up PRs now since we've grown quite a bit.

Thanks for catching this!

jef (approved):

PR #347, opened 2020-09-27
feat: retry logic for nvidia session token and adding to cart

andrewmackrodt:
The existing behaviour for obtaining a `session_token` and adding to cart assumes a best case scenario, i.e. nvidia's site is not overloaded and requests will succeed on the first try. We know that requests often timeout when there is a 3080/3090 drop so this PR aims to improve the entire process by optimizing when a session is retrieved and retrying the add to cart process.

**Note: all behaviour only applies if `OPEN_BROWSER=true`**.

1. The `session_token` is retrieved immediately as called by `generateSetupAction`, if this fails due to a gateway timeout error it is ok. A loop runs every `NVIDIA_SESSION_TTL` ms to renew the token, defaulting to 60 sec. Tokens should persist throughout the entire session but it's not clear how long they're valid for so we call the endpoint often to hint that we're still active.
2. When stock is available we should already have a `session_token`, saving us valuable milliseconds to add the product to our cart. If there is no `session_token` (extremely unlikely), we attempt to get one or throw an error. Finally we try to add the product to our cart up to `NVIDIA_ADD_TO_CART_ATTEMPTS` times, defaulting to 10. I've determined 10 to be a sane default value but we may want to increase it. Currently there is no delay between retrying requests, maybe this is something we want to support for large values.
3. Assuming 1 & 2 are successful, the cart redirect URL is opened in the user's browser as normal.


jef (approved):
Sweet! Thanks for bringing this back in.

jef:
> I've determined 10 to be a sane default value but we may want to increase it.

Some would say 10000 is sane 😆

PR #352, opened 2020-09-28
feat: support for proxy server

### LINKED ISSUE #321: Add support for a proxy or pool of IPs
### Description
Left image is a server on AWS. Right is powershell on my local computer. As far as I can tell, the AWS server's requests are blocked because the AWS region IP blocks are [easily discoverable](https://docs.aws.amazon.com/general/latest/gr/aws-ip-ranges.html).
image

### Possible solution

I'm sure there are other solutions, but adding support for a rotating block of proxied IP addresses would help reduce blocks. This can be done in the app directly by changing the IP address each time the browser session is started and stopped. If we want to change the IP address for each query, that would need to be implemented as a separate proxy server.

https://scrapoxy.io/
https://zenscrape.com/how-to-build-a-simple-proxy-rotator-in-node-js/

adrian549092:
### Description
Partially Fixes #321 
The proposed change allows the user to specify a `proxy server` ip address and port in the `.env` file to use for the connection to the internet. In certain networks internet access is restricted and is only allowed via a `proxy server`. 

### Testing
**Note:** Requires `proxy server` such as [squid-proxy](http://www.squid-cache.org/) to be able to test.
Once a `proxy server` is available, add the following entries to the `.env` file:
```
PROXY_ADDRESS="<IP ADDRESS OF PROXY SERVER>"
PROXY_PORT=<TCP PORT FOR PROXY SERVER>
```

Finally, if a values is provided for `PROXY_ADDRESS` and a connection to that address is not possible, the connections will fail with a message similar to the one below
```
[9:04:30 PM] error :: ✖ [nvidia] nvidia founders edition - net::ERR_PROXY_CONNECTION_FAILED at https://www.nvidia.com/en-us/shop/geforce/gpu/?page=1&limit=9&locale=en-us&categ
ory=GPU&gpu=RTX%203080

```


jef:
Nice, this will be good for those running in the cloud a well!

jef (approved):

superman500:
Can you add the fields for proxy authentication? or a way to feed it a list of proxies via command line like .........................................npm run start --proxy /users/user1/proxylist.txt 

then some format for the proxy files for it to use the proxies like http://username:password@61.18.233.35:60121

PR #354, opened 2020-09-28
Adds support for .env config path supplied via cmd line

IonCaza:
### Description

Preloading dotenv allows us to specify a .env config path on start such as: `npm start dotenv_config_path=./.env.normal`

This is useful for those of us that want to run multiple instances of the snatcher in the same folder.

Not supplying a dotenv_config_path makes the script use default value of .env in root.

### Testing

Tested with and without dotenv_config_path.


jef:
We should also change the way this gets imported.

PR #356, opened 2020-09-28
fix: corrected norway locales

AugustinWinther:
### Description

norway currency should be `NOK` not `EUR`
norway drLocale should be `no_no` not `nb_no`

_3080 example:_
Previous API url that nvidia.ts would call ended up beeing: 
https://api-prod.nvidia.com/direct-sales-shop/DR/products/nb_no/EUR/5438797200
this URL ends in **status: "failure"**

New fixed API url that nvidia.ts is calling: 
https://api-prod.nvidia.com/direct-sales-shop/DR/products/no_no/NOK/5438797200
This URL is correct

### Testing

Tested with no errors


jef (approved):
Thanks for the correction!

AugustinWinther:
> 
> 
> Thanks for the correction!

No problem! Happy to help :D

PR #366, opened 2020-09-28
chore: normalize ftw3 naming

pntless:
3090 ftw3 and ftw3 ultra were listed as ftw3 gaming and ftw3 ultra gaming. Removed gaming to bring them inline with the model name used everywhere else (as far as I can find).


jef:
Awesome! Thank you!

PR #374, opened 2020-09-29
(yet another) Docker image

tom300z:
### Description
I just noticed that im not the first one to submit a docker image (or the respective dockerfiles), so feel free to close the pull request if you think another approach suits the project better.

My docker image downloads and runs the latest nvidia snatcher automatically so maintenance should be fairly low.
This pull request basically only contains a reference to the image and my github so you wont have to build/maintain the image yourself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ You do not need any computer skills, smarts, or anything of that nature. You are
 
 ### Quick overview
 
+#### Native installation (recommended)
+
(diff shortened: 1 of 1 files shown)


jef:
I actually don't mind your minimal Dockerfile. I wouldn't mind having it being pulled into here if you'd like it here as well!

I can build out any ghcr pipeline around it as well.

Otherwise, looks good!

Issue #310, opened 2020-09-25
Let us set price ranges to avoid 3rd party sellers/price gouging


chineeboy:
### Description

Let us define a min and max price range so we can avoid price gougers and 3rd party sellers.  We'll only get notified when stock is available and falls within our specified price range.

### Possible solution

scanning the price tag and only notify if it falls within the price range specified in .env
MIN_PRICE="699.99"
MAX_PRICE="1199.99"

wedge22:
I just suggested the same implementation in another issues post, I fully support this idea.

xninjax:
Yes this would be very beneficial in combating scalpers.

geman220:
Is this only happening on Amazon?  The reason I ask is due to the fact that implementation is a bit more complex than just adding a min/max price filter.  For example, a 3070 would have a different min/max than a 3090.  I mean, we can make it so everything has different min/max configurations but that's a lot of rework if the problem is only Amazon.  **If** the problem is only happening on Amazon we could try to re-implement the "banned seller" feature which will disallow positives from Sellers who are known to scalp on Amazon.

chineeboy:
> Is this only happening on Amazon? The reason I ask is due to the fact that implementation is a bit more complex than just adding a min/max price filter. For example, a 3070 would have a different min/max than a 3090. I mean, we can make it so everything has different min/max configurations but that's a lot of rework if the problem is only Amazon. **If** the problem is only happening on Amazon we could try to re-implement the "banned seller" feature which will disallow positives from Sellers who are known to scalp on Amazon.

I can only speak for Amazon.ca and newegg.ca, they both allow 3rd party sellers

geman220:
I wasn't aware Newegg allowed 3rd party sellers.  I still think it might be easier to disallow sellers that you can add in `.env` since again, price ranges would be hard to implement across 3 different cards with different price ranges, although not impossible.

chineeboy:
> I wasn't aware Newegg allowed 3rd party sellers. I still think it might be easier to disallow sellers that you can add in `.env` since again, price ranges would be hard to implement across 3 different cards with different price ranges, although not impossible.

what if we had prices for each card? and to be honest, max price would be useful, min price less so.

something like max_price_2090, max_price_3080, max_price_3070.?

geman220:
[...]

PR #383, opened 2020-09-29
feat: max price filtering

xninjax:
### Description

This feature will address issues such as:  Let us set price ranges to avoid 3rd party sellers/price gouging #310

Improves filtering by allowing a supplied `MAX_PRICE` in .env.  This can help filter out price gouging sellers and is an alternate to having to specifically list banned sellers.

It would be better to have an array of prices (one for each card type) - but this should suffice as a framework to build upon.

### Testing

test:series card was used against all stores (updated some test card links for euro stores, to ensure euro price formatting was taken care of properly).  

Ensured that euro price formatting i.e 1.234,99 was converted to 'standard' formatting i.e 1234.99.  
- Tested against multiple amazon-de/nl products with varying number of digits in the pricetag, to ensure conversion works over a wide range of supplied prices
- store model must include optional `euroFormat: true` for conversion to take place (see example below)

Ensured that Max Price was obeyed
- MAX_PRICE="5"  was set in .env and all test cards were confirmed to be skipped over.  A new logger.info message is produced anytime the store price is greater than the threshold price.
- MAX_PRICE="" and MAX_PRICE="2000" was set in .env, and all test cards were again able to trigger in-stock alerts.

Note: Manufacturer direct sales sites (ASUS, NVIDIA, etc) do not have max_price selectors currently set - and is likely not required.

Reproducing the test is pretty easy, set a MAX_PRICE in .env to a low enough value.  Enable test:series card.  Output will display a logger.info message including the store card price vs threshold price, if test card price exceeds max price .

Optional Max price container must be set in store model, under 'labels', for price filtering to function for that store.
```
maxPrice: {
		container: 'span[class*="PriceString"]',
		euroFormat: true
	  }
```
euroFormat is optional, and only required when store provides euro price formatting - as described above.

### Functional demonstration
**Test 1**
`MAX_PRICE="0.1"`
image
✅  amazon-de and amazon-nl demonstrate proper currency format conversion (converts to dollars) and price limit filtering activated
✅  newegg-ca demonstrates proper currency format checking (remains as dollars) and price limit filtering activated 
✅  ASUS demonstrates ability to trigger alert due to no price container set in asus store model 

**Test 2**
`MAX_PRICE="325"`
image
✅  newegg-ca and amazon-de price remains above MAX_PRICE, and price limit filtering is activated
✅  amazon-nl  is now below MAX_PRICE and thus allowed to trigger alert
✅  ASUS same as before


xninjax:
@jef sorry for all the commits in this branch...  I'm definitely **not** a git-guru by any means ¯\\\_(ツ)\_/¯

Hope it wasn't too early for a PR, as testing on other pages hadn't been done (aside from amazon/-ca and newegg/-ca), but the framework is there now.  My intent of max price should really only apply to sites with 3rd party  marketplace sellers anyhow...  I wouldn't imagine a need for max price on the manufactures sites?

karmanbadhesha:
I had accomplished something similar by looking at the "Sold and shipped by Amazon.ca" but it can result in cases where a cheap 3rd party reseller listing might get missed! Price is definitely the way to go

src/store/lookup.ts
```diff
@@ -145,6 +145,35 @@ async function lookupCardInStock(store: Store, page: Page, link: Link) {
 		}
 	}
 
+	if (store.labels.maxPrice && Config.store.maxPrice) {
```
jef, line 148: Do you mind abstracting this logic in this if block just like the if statements above and below using the `pageIncludesLabels`?

It would help with complexity.

That being said, take a look at that function. I believe what you're wanting is something very similar. As I think your approach is great, but I think there is some extra work you're doing here that may not be needed since we have those helper functions.

src/store/model/adorama.ts
```diff
@@ -9,6 +9,10 @@ export const Adorama: Store = {
 		inStock: {
 			container: '.buy-section.purchase',
 			text: ['add to cart']
+		},
+		maxPrice: {
+			container: '.your-price',
```
jef, line 14: Really? `your-price` is their class? Who's in charge of their naming 😆

jef (changes requested):
Overall, I really like it! I appreciate the contribution! I know a lot of people will like this. Especially with Amazon third party sellers.

Just a few comments. Thanks for all the help!

src/store/lookup.ts
xninjax: Yes no worries, I'll take a look, appreciate the feedback! :sunglasses:

jef: Thanks! Look forward to see what's to come.

xninjax: Abstraction injection sequence complete 😄 
- Could not utilize `pageIncludesLabels` directly - as a mathematical comparison was needed, but it became a good jumping off point as reference for a new helper function...  
 Open to feedback 👍

xninjax:
Good question about the containers, what I was finding was that the price returned from the original/simple selectors were sometimes returning a completely different value... It appeared some js would modify prices and containers after load, likely bot protection 🤖  The more complex container matching helped ensure to grab the final result for the actual price 🏆 😁

HickTrick:
Hey there, I was just testing this out against amazon.ca where the TUF OC has a scalper trying to sell it for $1,879.99 CDN and it wasn't working.
This is what a price span looks like on amazon.ca:
`<span id="priceblock_ourprice" class="a-size-medium a-color-price priceBlockBuyingPriceString">CDN$&nbsp;1,879.99</span>`

I looked into it, and in the `cardPriceLimit` function, at the return statement, the `cardpriceNumber` is showing as `1.879` instead of `1879`

xninjax:
> 
> 
> Hey there, I was just testing this out against amazon.ca where the TUF OC has a scalper trying to sell it for $1,879.99 CDN and it wasn't working.
> This is what a price span looks like on amazon.ca:
> `<span id="priceblock_ourprice" class="a-size-medium a-color-price priceBlockBuyingPriceString">CDN$&nbsp;1,879.99</span>`
> 
> I looked into it, and in the `cardPriceLimit` function, at the return statement, the `cardpriceNumber` is showing as `1.879` instead of `1879`

Thanks for letting me know about this! - I'll take a look.  Perhaps the regex parsing needs some tweaking.  Do you have a link to the item still?  

As for the price container - for amazon currently using `span[class*="PriceString"]'` as a selector.  It has wildcard matching, as sometimes the id changes to something like `span id="priceblock_saleprice"` while other times it's like what you posted here.  The selector seems to have picked the correct container, but regex is not parsing the decimal place correctly now.

HickTrick:
> > Hey there, I was just testing this out against amazon.ca where the TUF OC has a scalper trying to sell it for $1,879.99 CDN and it wasn't working.
> > This is what a price span looks like on amazon.ca:
> > `<span id="priceblock_ourprice" class="a-size-medium a-color-price priceBlockBuyingPriceString">CDN$&nbsp;1,879.99</span>`
> > I looked into it, and in the `cardPriceLimit` function, at the return statement, the `cardpriceNumber` is showing as `1.879` instead of `1879`
> 
> Thanks for letting me know about this! - I'll take a look. Perhaps the regex parsing needs some tweaking. Do you have a link to the item still? EDIT: I found the card on amazon. will test.
> 
> As for the price container - amazon currently using `span[class*="PriceString"]'` as a selector. It has wildcard matching, as sometimes the id changes to something like `span id="priceblock_saleprice"` while other times it's like what you posted here. The selector seems to have picked the correct container, but regex is not parsing the decimal place correctly now.

Glad you found the card, the scalper has been putting it up and taking it down repeatedly, so it’s not always there with that price. 

And yes, I was using the selector as you had it with the wildcard matching. I threw in some log statements and saw that it was grabbing that whole string of `CDN$&nbsp;1,879.99`. I didn’t break apart your one line that did the regex matching and joining after to see exactly where this was occurring. 

It’s possible I messed something up, as I more or less manually copied your changes into my locally cloned master, for some reason git wouldn’t let me pull in your forked branch. So maybe I missed something when I was doing that, not sure.

xninjax:
> Glad you found the card, the scalper has been putting it up and taking it down repeatedly, so it’s not always there with that price.
> 
> And yes, I was using the selector as you had it with the wildcard matching. I threw in some log statements and saw that it was grabbing that whole string of `CDN$&nbsp;1,879.99`. I didn’t break apart your one line that did the regex matching and joining after to see exactly where this was occurring.
> 
> It’s possible I messed something up, as I more or less manually copied your changes into my locally cloned master, for some reason git wouldn’t let me pull in your forked branch. So maybe I missed something when I was doing that, not sure.

Ok, found the issue. 
After I changed to passing the regex as a variable into the parser (instead of hard-coded), I must have missed re-testing with prices that contains commas (or contains periods for euro format). 🤦‍♂️ 

By removing the quotations around the regex stored into the `priceSeperator` variable, everything began to work again.

Should be: `const priceSeperator = query.euroFormat ? /\./g : /,/g;`  _<note no quotes around regex_

Thank-you @HickTrick for pointing this out - big save!

(incoming commit soon)

**Test on the amazon-ca scalper card**
image

**Test on Euro laptop**
image

HickTrick:
> Ok, found the issue.
> 
> (incoming commit soon)

Awesome, glad you found it and figured out the solution! Glad to help!

jef (approved):
Very thorough work done here. I _really_ appreciate the time spent and the considerations you made. Thanks for making this project better!

PR #384, opened 2020-09-29
feat: Added GameStop as a store

Agentscreech:
### Description

Added a new store, GameStop.  They only have one card listed right now

### Testing

Validated that it hits the endpoint. Reports currently out of stock correctly.

### DIFF
--- a/src/store/model/gamestop.ts
+++ b/src/store/model/gamestop.ts
@@ -1,0 +1,19 @@
+import {Store} from './store';
+
+export const Gamestop: Store = {
+	labels: {
+		inStock: {
+			container: '#primary-details .primary-details-row .cart-and-ipay.divider-line.no-border-mobile',
+			text: ['ADD TO CART']
+		}
+	},
+	links: [
+		{
+			brand: 'asus',
+			model: 'tuf oc',
+			series: '3080',
+			url: 'https://www.gamestop.com/video-games/pc/components/graphics-cards/products/tuf-gaming-geforce-rtx-3080-graphics-card/11109446.html'
+		}
+	],
+	name: 'gamestop'
+};
--- a/README.md
+++ b/README.md
@@ -30,11 +30,11 @@ The purpose of this bot is to get an Nvidia card. It tries multiple things to do
 
 > :point_right: You may get false positives from time to time, so I apologize for that. The library currently waits for all calls to be completed before parsing, but sometimes this can have unknown behavior. Patience is a virtue :)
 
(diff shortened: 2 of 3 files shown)


Agentscreech:
other than the lint error. looks like the element they use, I would want to look at `.textContent` for the button in question. However, the code for `inStock` checking tramples over the default of `.textContent` and uses `.outerHTML`. Might have to look in to a way to have that set in the store config.

jef:
A good way to avoid getting the divs/containers trampled is perhaps grabbing the cart element's parent, or even it's parent.

src/store/model/gamestop.ts
```diff
@@ -0,0 +1,19 @@
+import {Store} from './store';
+
+export const Gamestop: Store = {
+	labels: {
+		inStock: {
+			container: '#primary-details .primary-details-row .cart-and-ipay.divider-line.no-border-mobile',
```
jef, line 6: You may just want to use one container or class that is relevant.

```suggestion
			container: '.add-to-cart-buttons',
```

`.add-to-cart-buttons` seems like a good candidate. It's on other stocked items as well.

image

Agentscreech: yeah, I think just `.add-to-cart` works in newer testing. Still doesn't solve the issue with the outerHTML being forced on the store.lables.inStock test.

src/store/model/gamestop.ts
```diff
@@ -0,0 +1,19 @@
...
+			text: ['ADD TO CART']
```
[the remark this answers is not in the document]
Agentscreech: So even though the site will do all caps, we can just put all lower case then?

src/store/model/gamestop.ts
Agentscreech: This would use `textContent`, but not sure how to code with with all the strict typing things that are going on. Never used typescript before.

src/store/model/gamestop.ts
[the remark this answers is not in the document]
jef: Yep, we convert it in `includesLabels()`

https://github.com/jef/nvidia-snatcher/blob/b868d1a4833a8ec5ac1c79481530d75cd0c4b01e/src/store/includes-labels.ts#L89-L92

jef:
This ended up getting merged in: https://github.com/jef/nvidia-snatcher/pull/390

This was a regression from it: https://github.com/jef/nvidia-snatcher/issues/392

jef:
Thanks for your contribution!

src/store/model/gamestop.ts
jef: No worries! Yeah in this case, you don't need to have the typings within this scope. We just put in the element.

PR #389, opened 2020-09-30
feat(store): add french store topachat

oyagci:
### Description

Add one of the most popular french store for gaming products.

### Testing

I've tested the matches with products that are already in stock (like a GTX 1660)


jef:
Looking into. Need to verify everything works good!

Thanks for the PR.

jef:
Thanks @oyagci! I've made another PR to fix this up and included you part of that commit history.

PR #403, opened 2020-10-01
feat: Add's a default sound to .env_example and bundles the required audio file

afisherdev:
feat: Add's a default sound to .env_example and bundles the required audio file

### Description

I included a default path for audio alerts as well as bundled the audio file so people are aware of where they should place their audio files for audio alerts. I also bundled an audio file created for users to motivate them to buy as fast as possible. 

### Testing

I ran the included test:notification script included in the package.json and it works as expected.

### New dependencies

None


jef:
As much as I think this could be helpful to users, I also think it doesn't matter.

The default location for a sound file is the root of the project. Putting it in another folder makes this equally as "complicated".

I also don't like storing binaries in GitHub (see https://github.com/jef/nvidia-snatcher/pull/52#pullrequestreview-491997101).

I'd like to keep the git fetch as minimal as possible and putting random sounds in the history defeats that.

Thanks for the contribution and your time though! 😀

PR #404, opened 2020-10-01
fix: add status code `429` to bandh `backOffStatusCodes`

adolfox:
### Description

processBackoffDelay only triggers with status codes 403.

I was getting a lot of RATE LIMIT EXCEEDEDs from B&H.  
bandh_rate_limit_exceeded

In lookupCard, however, it does check for 429 and logs it as a warning.
```
if (statusCode === 429) {
			Logger.warn(Print.rateLimit(link, store, true));
```

I figure that getting these RATE LIMIT EXCEEDEDs is bad and it should trigger a backoff, so I added 429 to bandh.ts to the backoffStatusCodes: [403, 429] array/list.

A lot less RATE LIMIT EXCEEDEDs
bandh_rate_limit_exceeded_backoff_working

### Testing

I've been running it with the added 429 and it's been running fine for the last few hours with no issues.  Less RATE LIMIT EXCEEDEDs.

edit: code formatting

### DIFF
--- a/src/store/model/bandh.ts
+++ b/src/store/model/bandh.ts
@@ -1,6 +1,7 @@
 import {Store} from './store';
 
 export const BAndH: Store = {
+	backoffStatusCodes: [403, 429],
 	labels: {
 		inStock: {
 			container: 'div[data-selenium="addToCartSection"]',


jef (approved):
Great! Sounds good to me. Weird that we were getting 403s... I guess we get blocked for a certain amount of time.

Thanks for bringing this one in.

PR #419, opened 2020-10-02
feat(store): adds models to bestbuy

pntless:
### Description

Adds 3080 xc3, ftw3, ftw3 ultra, aorus master, vision and 3090 ftw3, ftw3 ultra
Adds vision to readme list of models


jef:
Any reason why you closed this out?

pntless:
I'm new to github, still figuring it out... I realized it sent like 8 commits with this PR instead of just the two I wanted. I resubmitted.

jef:
That's okay! They get squashed when I merge it in. Meaning, they get combined in all one in a pretty bow :)

pntless:
Ahh good to know. Thanks for the tip :)

PR #431, opened 2020-10-03
chore(store): update aorus master 3080 links

pntless:
Gigabyte rep on reddit provided an inaccurate link ahead of launch. I've updated it to reflect the actual live link now that the page is up.


jef (approved):
Thank you @pntless! :rocket:

PR #451, opened 2020-10-05
feat: max price per series

jastheace:
### Description

N/A - max price is for single product; this makes it by product

Updates max price so it is product specific instead of being a single number

Jase/Myanth#7701

### Testing

untested but works in theory; don't have an endpoint that breaks the price currently

changes maxPrice to maxPrice3070, maxPrice3080, maxPrice3090


src/config.ts
```diff
@@ -141,7 +141,9 @@ const proxy = {
 
 const store = {
 	country: envOrString(process.env.COUNTRY, 'usa'),
-	maxPrice: envOrNumber(process.env.MAX_PRICE),
+	maxPrice3070: envOrNumber(process.env.MAX_PRICE_3070),
+	maxPrice3080: envOrNumber(process.env.MAX_PRICE_3080),
+	maxPrice3090: envOrNumber(process.env.MAX_PRICE_3090),
```
jef, line 146: Users will need to know this is breaking change for `MAX_PRICE` though. Perhaps we should include a warning for now and then we can delete later.

```typescript
if (process.env.MAX_PRICE) logger.warn('ℹ MAX_PRICE is deprecated, please use MAX_PRICE_${series}')
```

jef (changes requested):
This makes a lot of sense to me. Just one comment on reporting to the user of breaking change should be sufficient.

Thanks!

src/config.ts
jastheace: added relevant section to index with comments so other deprecated items can be added in the pre-execution

jef (approved):
Great! Thank you very much for cleaning this up and adding the message.

Issue #370, opened 2020-09-28
Email/SMS notifications and a VPN


towerjunkie:
First of all, i want to thank you guys for making such a useful and simple software. 

The only real issue i have is that my notifications work fine when i'm on my actual machine but when i switch to a VPN, none of these notifications work. 
This is the error i receive: couldn't send email Invalid login: 534-5.7.14

I'm using Gmail and i already set it up to allow less trustworthy sign-ins.

gonkar193:
Odd, I have mine going through a VPN and everything is working perfectly fine. Are you sure you have your .env set up correctly?

towerjunkie:
Yeah. I have it working fine when I'm going through my regular machine. I
just get an error when using nord vpn.
Here is a screenshot of on and Off.

But listen, I'll deal with it. I don't want you to have to focus on this
issue if a solution is not on hand.

Again, thank you for creating this open software.

jxjftw:
Maybe the smtp relay is blocking the VPN external IP?

tempestbk:
Firstly, thank you for creating the wonderful piece of software. I've tried all the solutions and I still cannot get the program to notify me via email or sms. I've turned on less secure app access. I've tried plugging in the default smtp info for gmail. What am I doing wrong? any suggestions?

jef:
That's gotta be frustrating. Are you also running in a VPN?

tempestbk:
Hello Jef,

no problem somebody in your discord helped me out. We both don't know what
it was copied his config and everything worked.

P.S. I know you are busy is there any chance you will add more Canadian
stores to the project in the future? Like Canada Computers, Memory Express

jef:
I don't know all of those stores and rely a lot on contributors. But it's definitely a possibility!

tempestbk:
Where should I put the suggestion?

jef:
If you want to create a new issue there, that works great.

PR #364, opened 2020-09-28
fix: `envOrNumber` behavior

andrewmackrodt:
See #363 and #359

PR #452, opened 2020-10-05
feat: support for smtp server

adrian549092:
### Description
No issue associated with this feature but could be a solution for #370
The proposed change allows the user to specify an `smtp server` ip address or fqdn and port in the `.env` file to use for sending email and sms notifications. In my particular situation, the application resides behind a proxy server which does not allow connections to external smtp servers but I do have a local mail relay agent available for sending emails. 

### Testing
When `SMTP_ADDRESS` and `SMTP_PORT` are left unspecified the application will use the default `gmail` service and works as normal. When the mentioned variables are set then the application will use the specified settings to connect to an SMTP Relay Agent and send emails. Both of the previously mentioned cases were tested using the available `test:notification` `npm script` and I also set the `PHONE_NUMBER` and `PHONE_CARRIER` variables in order to confirm that SMS notifications worked as well. **All tests succeeded**.
```
SMTP_ADDRESS="<IP ADDRESS OF SMTP SERVER>"
SMTP_PORT=<TCP PORT FOR SMTP SERVER>
```


src/notification/email.ts
```diff
@@ -6,12 +6,23 @@ import nodemailer from 'nodemailer';
...
-		user: email.username
-	},
-	service: 'gmail'
+const transportOptions: any = {};
+
+if (email.username && (email.password || email.smtpAddress)) {
+	transportOptions.auth = {};
+	transportOptions.auth.user = email.username;
+	transportOptions.auth.pass = email.password;
+}
```
jef, line 15: It's OK if password is empty?

```suggestion
if (email.username && (email.password || email.smtpAddress)) {
	transportOptions.auth = {};
	transportOptions.auth.username = email.username;
	transportOptions.auth.password = email.password;
}
```

Name change for consistency purposes.

src/notification/email.ts
```diff
@@ -6,12 +6,23 @@ import nodemailer from 'nodemailer';
...
+
+if (email.smtpAddress) {
+	transportOptions.host = email.smtpAddress;
+	transportOptions.port = (email.smtpPort ? email.smtpPort : 25);
```
jef, line 19:
```suggestion
	transportOptions.port = email.smtpPort;
```

This is fine because of the defaulting done in `config.ts`.

src/config.ts
```diff
@@ -133,6 +135,11 @@ const page = {
 	width: 1920
 };
 
+const proxy = {
+	address: envOrString(process.env.PROXY_ADDRESS, ''),
```
jef, line 139:
```suggestion
	address: envOrString(process.env.PROXY_ADDRESS),
```

Default is empty.

jef (changes requested):

src/notification/email.ts
adrian549092: Yes, if an `smtpAddress` is configured, then a password is not required because some smtp relay agents don't require authentication (like the one I use). Also, `auth.user` and `auth.pass` are the attribute names that `createTransport` is expecting [nodemailer-Authentication](https://nodemailer.com/smtp/#authentication)  so changing them to `auth.username` and `auth.password` might break it.

jef: Makes sense to me then. No need to change.

src/notification/email.ts
jef: As long as your set your environment variable ot `SMTP_PORT=""` or removing that line, it should be returning `25` since you're using `envOrNumber(process.env.SMTP_PORT, 25)`.

jef:
Once conflicts are resolved, would love to have this in!

Thanks for working on this and the previous proxy work.

src/notification/email.ts
adrian549092: right, yea that is exactly what I would expect would happen, however I am getting a returned value of `0` for some reason... I just checked on a couple other Config items that also use `envOrNumber` and have a default specified and I am getting `0` for those as well..

jef: Interesting... I did the same and didn't get those results which is strange. What OS and are you running in terminal or IDE?

adrian549092: Wow, that is so weird. I am running Windows 10 on IDE (WebStorm)... Oh WAIT... I wonder if its because I am running Node 12.14 instead of 14...

andrewmackrodt: You'll want to update your branch with `main`, this was fixed in #364

adrian549092: Yea I think that's what the problem is. I went ahead and commited the recommendation.

adrian549092:
no problem @jef my pleasure. let me know if there is any other action items on my end

adrian549092:
@jef and @andrewmackrodt I am going to close this PR and submit a new one as I am noticing that there are alot of conflicts not related to my feature. I will sync my fork then re-add the feature code and generate a new PR

PR #455, opened 2020-10-05
feat: add uk stores

andrewmackrodt:
**Relates to #461 - merge second**

This adds 10 UK stores, I've not added `maxPrice` yet but plan to do so in a future PR. https://github.com/andrewmackrodt/nvidia-snatcher/compare/pr-framework-updates...pr-uk-stores

### DIFF
--- a/src/store/model/amazon-uk.ts
+++ b/src/store/model/amazon-uk.ts
@@ -1,0 +1,95 @@
+import {Link, Store} from './store';
+import {logger} from '../../logger';
+import {parseCard} from './helpers/card';
+
+export const AmazonUk: Store = {
+	backoffStatusCodes: [403, 429, 503],
+	labels: {
+		captcha: {
+			container: 'body',
+			text: ['enter the characters you see below']
+		},
+		inStock: {
+			container: '#availability',
+			text: ['in stock']
+		},
+		maxPrice: {
+			container: 'span[class*="PriceString"]'
+		},
+		outOfStock: [
+			{
+				container: '#availability',
+				text: ['out of stock', 'unavailable']
+			},
+			{
+				container: '#backInStock',
+				text: ['unavailable']
+			}
+		]
+	},
+	links: [
+		{
+			brand: 'test:brand',
+			cartUrl: 'https://www.amazon.co.uk/gp/aws/cart/add.html?ASIN.1=B081265T5Z&Quantity.1=1',
+			model: 'test:model',
+			series: 'test:series',
+			url: 'https://www.amazon.co.uk/dp/B081265T5Z/'
+		}
+	],
+	linksBuilder: {
+		builder: (docElement, series) => {
+			const productElements = docElement.find('.s-result-list .s-result-item[data-asin]');
+			const links: Link[] = [];
+			for (let i = 0; i < productElements.length; i++) {
+				const productElement = productElements.eq(i);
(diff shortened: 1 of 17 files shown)


src/store/model/helpers/card.ts
```diff
@@ -1,13 +1,91 @@
...
+	};
+	/* eslint-enable unicorn/no-fn-reference-in-iterator */
+}
+
 export function parseCard(name: string): Card | null {
-	name = name.replace(/[^\w ]+/g, '').trim();
-	name = name.replace(/\bgraphics card\b/gi, '').trim();
-	name = name.replace(/\b\w+ fan\b/gi, '').trim();
-	name = name.replace(/\s{2,}/g, ' ');
+	name = name.replace(/\w+-\w+-[^ ]+/g, '');
```
jef, line 75: Wow 😮 Nice work!

src/store/model/amazon-uk.ts
```diff
@@ -0,0 +1,95 @@
...
+	links: [
+		{
+			brand: 'test:brand',
+			cartUrl: 'https://www.amazon.co.uk/gp/aws/cart/add.html?ASIN.1=B081265T5Z&Quantity.1=1',
+			model: 'test:model',
+			series: 'test:series',
+			url: 'https://www.amazon.co.uk/dp/B081265T5Z/'
+		}
+	],
+	linksBuilder: {
```
jef, line 39: I imagine we could use this for the other Amazon stores, but could be very wrong there.

jef (approved):
Wowa wowa. It looks great. Thanks for putting in this work. You're on the hunt for that card 😎

jef:
> We can break this into 2 different PRs or commits squashed into the 2 units once approved.

No problem. I don't mind it in both.

andrewmackrodt:
> You're on the hunt for that card sunglasses

I actually snagged a 3080 the same night/morning nvidia made the API changes but trying to do what little I can to help others get their cards :)

jef:
Well we all appreciate your contribution ❤️

src/store/model/amazon-uk.ts
andrewmackrodt: I imagine the `builder` function would work between stores, but determining the `urls[].url[]` would require a manual map per region. At a minimum, the following may work (not tested it compiles let alone works):

```ts
function getAmazonQueryString(search: string, categoryId: number, includeOutOfStockId: number) {
    return [
        `k=${encodeURIComponent(search)}`,
        'rh=' + encodeURIComponent([
            `n:${categoryId}`,
            `p_n_availability:${includeOutOfStockId}`
        ].join(','))
    ].join('&');
}

const queryString = getAmazonQueryString('+"RTX 3080" -2080 -GTX', 430500031, 419162031);
const domain = 'www.amazon.co.uk';
const url = `https://${domain}/s?${queryString}`;
```

andrewmackrodt:
@jef I've split the first changset into #461 and will rebase this PR on top of main once that one's merged.

src/store/model/amazon-uk.ts
jef: That makes sense to me. Maybe we can bring this in at a later point. Thanks for checking this out.

jef (approved):
Thank you very much for all this hard work!

PR #458, opened 2020-10-05
feat: add support for specifying smtp server

adrian549092:
### Description
No issue associated with this feature but could be a solution for #370
The proposed change allows the user to specify an `smtp server` ip address or fqdn and port in the `.env` file to use for sending email and sms notifications. In my particular situation, the application resides behind a proxy server which does not allow connections to external smtp servers but I do have a local mail relay agent available for sending emails. 

### Testing
When `SMTP_ADDRESS` and `SMTP_PORT` are left unspecified the application will use the default `gmail` service and works as normal. When the mentioned variables are set then the application will use the specified settings to connect to an SMTP Relay Agent and send emails. Both of the previously mentioned cases were tested using the available `test:notification` `npm script` and I also set the `PHONE_NUMBER` and `PHONE_CARRIER` variables in order to confirm that SMS notifications worked as well. **All tests succeeded**.
```
SMTP_ADDRESS="<IP ADDRESS OF SMTP SERVER>"
SMTP_PORT=<TCP PORT FOR SMTP SERVER>
```


jef (approved):
Thanks for taking care of the conflicts. It was quite a bit :sweat_smile:

:shipit:

PR #461, opened 2020-10-06
feat: framework updates

andrewmackrodt:
**Relates to #455 - merge first**

I began the concept of a `linksBuilder` in #270, this allows us to detect links for a store without having to maintain a hardcoded list of URLs. This expands upon the part of work by introducing an optional `ttl` (the time in milliseconds to check for new links), adds a generic helper method to extract links from several stores, and updates the helper method to parse a card's store title into our model and brand names. For an example of why this is useful, see the huge number of listings Amazon have: https://github.com/jef/nvidia-snatcher/issues/357#issuecomment-703589117.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,6 @@
 import {Stores} from './store/model';
 import {adBlocker} from './adblocker';
 import {config} from './config';
-import {fetchLinks} from './store/fetch-links';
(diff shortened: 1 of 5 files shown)


andrewmackrodt:
Closing as all changes already part of #455 .

PR #465, opened 2020-10-07
bugfix: env min/max input validation

xninjax:
### Description
_This was originally going to be a bug report, instead submitting PR with possible solution._

**ISSUE**
When supplying **only one** of a `min`/`max`  pair from .env, undesirable results can occur in code.
I discovered this when setting _only_ `PAGE_SLEEP_MIN="60000"`  ... resulting in the page hits to occur more often than 60sec.

image
image

Why?  
The culprit 
` getSleepTime() {
	return config.browser.minSleep + (Math.random() * (config.browser.maxSleep - config.browser.minSleep));
}`

Since `PAGE_SLEEP_MAX` defaults to `"10000"`

` getSleepTime() {
	return 60000 + (Math.random() * (10000 - 60000));
}`
                                                                   
We're applying a negative number to the Math.random result ☝️ .  Thus subtracting from the original 60sec delay.

**Possible solution**

`... Math.random() * Math.abs(config.browser.maxSleep - config.browser.minSleep)) ...`

`Math.abs...` is one possible fix,  but doesn't really get the job done properly IMO. Would still need to think about how to address this for `PAGE_BACKOFF_MIN/MAX`.  And again for all future min/max pairs.

Instead, I propose `MIN/MAX` input validation (subject of this PR). 
```
function envOrNumberMinMax
  Returns environment variable, given number, or default number,
  while handling .env input errors for a Min/Max pair.
  .env errors handled:
  - Min/Max swapped (Min larger than Max, Max smaller than Min)
  - Min larger than default Max when no Max defined
  - Max smaller than default Min when no Min defined
```

### Testing
Input validation applied...
With `PAGE_SLEEP_MIN="60000"` entered as before, and appling input validation, results are now as expected.
image
image

Other Scenario's
image
image

The following works because `input validation` swaps min/max first (not well illustrated, but was verified)
image
image


xninjax:
linter is not liking the single line 'if' statements, and I knew it would complain, since local checks warned me too.

I can expand, but will look more of a mess 🤷‍♂️

jef:
> `getSleepTime() { return 60000 + (Math.random() * (10000 - 60000)); }`

Ahhhhhh... Good find! Definitely want to bring this in. I can help with the linting.

src/config.ts
```diff
@@ -48,14 +50,66 @@ function envOrNumber(environment: string | undefined, number?: number): number {
...
+					return Number(Number(enviromentMin) < Number(enviromentMax) ? enviromentMin : enviromentMax);
+				}
+
+				if (enviromentMax) {
+					return Number(enviromentMax) < (number ?? 0) ? Number(enviromentMax) : (number ?? 0);
+				}
+
+				if (enviromentMin) {
+					return Number(enviromentMin);
+				}
```
xninjax, line 80: All expanded so linter is happy 😄

jef (approved):
Great! Thank you very much for doing this. I really appreciate it. This will hopefully clear up some issues.

PR #472, opened 2020-10-07
fix: trim strings from comma-separated values

er1992:
Trim strings from comma-separated values


jef (approved):
Cool! Makes sense to me. Thanks for doing this 😃

PR #486, opened 2020-10-09
chore: add typescript definitions for custom types

andrewmackrodt:
I've gone through the `README` of imported modules without an `@types/package` definition and added the basic API usage.


src/types/pushbullet.d.ts
```diff
@@ -1 +1,59 @@
-declare module '@hijef/pushbullet';
+declare module '@hijef/pushbullet' {
```
jef, line 1: Sweet! I'll put this into the module as well.

jef (approved):
Dang, this is a nice QoL change! Thanks for putting this in!

PR #489, opened 2020-10-10
feat(store): add rog-strix-3080 to newegg-ca

xninjax:
### Description
ROG Strix 3080 is listed on newegg-ca now.  So adding it to store model.
Checked other store models (bestbuys, newgg, etc) - but only newegg-ca seems to need the update (so far).

yarrpirate#8807

### Testing
 no warnings generated, only oos message when checking against new url


jef (approved):

jef:
Sweet! Thanks @xninjax ❤️

PR #513, opened 2020-10-13
fix: dutch stores components

FeikoWielsma:
### Description
This should fix some items (MSI Ventus and Trio) always showing in stock in AmazonNL.
I also added 3090 cards to Coolblue, and some more 3080's to AmazonNL, along with some small model fixes.

### Testing
I tested one some other cards in various states of stock, seemed to respond normally


jef:
Awesome. Thanks for doing this!

PR #521, opened 2020-10-14
feat: add caseking and proshop-de store

TaliZorahVasNormandy:
### Description

Added German Caseking and Proshop Store with 3080 and 3090 models


jef (approved):
Looks good to me! Thanks for the great addition 😄

PR #524, opened 2020-10-15
feat: add proshop-dk

TaliZorahVasNormandy:
### Description
Added Danish Proshop


jef (approved):
Thanks! This is great.

PR #526, opened 2020-10-15
fix: add support for stores label text to have uppercase letters

farsabbutt:
### Description
Currently the labels text to be searched must only be in lowercase for the search in DOM text to work successfully. If the provided labels text has uppercase letters, it will not be matched successfully. 

Some contributors might not know about this restriction and may set text containing uppercase letters.

With this commit, It should be possible to have uppercase letters in the labels text. This change should be backwards compatible, and should work for both cases with labels having lower case letters or upper case letters.

### Testing
If you set some label text that contains any uppercase letter, it will not be found in the DOM text because the DOM text is being lowercased before search and match while the label text is not lowercased. With this change we lowercase the label text too before performing search and match, hence both uppercase letters and lowercase letters work fine.

### DIFF
--- a/src/store/includes-labels.ts
+++ b/src/store/includes-labels.ts
@@ -86,7 +86,7 @@ export async function extractPageContents(page: Page, selector: Selector): Promi
  */
 export function includesLabels(domText: string, searchLabels: string[]): boolean {
 	const domTextLowerCase = domText.toLowerCase();
-	return searchLabels.some(label => domTextLowerCase.includes(label));
+	return searchLabels.some(label => domTextLowerCase.includes(label.toLowerCase()));
 }
 
 export async function cardPrice(page: Page, query: Pricing, max: number, options: Selector) {


jef:
I like and don't like this. It enables the developer to be inconsistent. But for those who are updating labels locally and not submitting PRs, I suppose could be helpful. I'm hesitant to put this in.

I think there is a better chance of this getting merged if you were able to get the labels lowercased before these get evaluated (maybe in `store/index.ts`?).

farsabbutt:
I was actually adding a new store and copied the 'Add to cart' text directly from the webpage of the store and it did not work. I did not know that it has to be lowercase. So I had to spend some time to figure out what was going on, since the 'Add to cart' was displayed on the page, I expected it to trigger the notification but it did not. 

That is why I thought others might have the same issue and its just a suggestion to prevent them from having the same issue :)

Yes I agree with you, It can cause inconsistency, but then a clear way to indicate that the labels must be lower case only would be nice too.

jef:
Alright, let's go ahead and merge this in just so people don't waste their time trying to debug their sound logic.

Will create a ticket later on. Thanks!

jef (approved):

PR #528, opened 2020-10-15
feat: twitch chat notification, multiple telegram chat ids and german web shop overhaul

### LINKED ISSUE #530: Twitch support?
add twitch support?

I was going to buy a dedicated server and figured I would stream it and it would be nice to post the link on twitch chat when something comes in stock

PaddseL:
### Description

Resolves #530 

feat: allow multiple telegram chats to be notified
feat: add notifications to twitch channel chats

feat: add computeruniverse (de)
feat: add cyberport (de) shop
feat: add mindfactory (de) shop
feat: add saturn (de) shop (see @fubu2k's [pr](https://github.com/jef/nvidia-snatcher/pull/494))

chore: normalize model names for various german stores
chore: add and sort links for various german stores
chore: removed dead links from various german stores
chore: shortened links for proshop-de

sorry for this huge pr, i somehow forgot to create one after each feature as i was constantly adding and testing.

### Testing
telegram notifications work this way for either a single chat id or comma seperated chat ids,
as `telegram.chatId.split(',')` returns the original string if no comma is found.

tested twitch notifications and found this way to be most stable. first i tried to only connect in an event of an alert and
immediately disconnect when the message got sent, but this introduced some weird behavior and the `ChatClient` with
the `RefreshableAuthProvider` should maintain the connection or reconnect in case of an connection loss.
[the setup of this notification method is a bit laborious though](https://d-fischer.github.io/twitch-chat-client/docs/examples/basic-bot.html), maybe this should be mentioned in the README.
i'm open for discussion on this one and would like to hear your opinion.

i found it to be more reliable to check for `outOfStock` instead of `inStock` for most german webshops, as they rarely list new items as "in stock" rather than more something like "available in x days" and accept preorders, so they don't even get to the "in stock" state.

### New dependencies
`"twitch": "^4.2.6"`,
`"twitch-auth": "^4.2.6"`,
`"twitch-chat-client": "^4.2.6"`


johnconnor2020:
Incredible work. Thank you.

johnconnor2020 (approved):

Floppy012 (dismissed):
Looking good to me :+1:

jef:
Holy moly. I'll give this a review today and we can discuss if necessary. Thanks!

src/notification/telegram.ts
```diff
@@ -15,13 +15,18 @@ export function sendTelegramMessage(link: Link, store: Store) {
...
 		(async () => {
 			const givenUrl = link.cartUrl ? link.cartUrl : link.url;
+			const results = [];
 
-			try {
-				await client.sendMessage(telegram.chatId, `${Print.inStock(link, store)}\n${givenUrl}`);
-				logger.info('✔ telegram message sent');
-			} catch (error) {
-				logger.error('✖ couldn\'t send telegram message', error);
+			for (const chatId of telegram.chatId.split(',')) {
```
jef, line 20: I'd add a note in the telegram doc in `README`. I'd also update `config.ts` to use `envOrArray()`.

src/notification/twitch.ts
```diff
@@ -0,0 +1,54 @@
...
+					refreshToken
+				}, null, 4), 'utf-8');
+			},
+			refreshToken: tokenData.refreshToken
+		}
+	),
+	{
+		channels: [twitch.channel]
+	}
+);
```
jef, line 39: I believe this will cause a problem if the user does no have any Twitch auth setup. Perhaps wrap this around a check.

src/store/model/alternate.ts
```diff
@@ -2,10 +2,6 @@ import {Store} from './store';
 
 export const Alternate: Store = {
 	labels: {
-		inStock: {
```
jef, line 5: How come you removed this label? If it's the incorrect label, we should update it. Otherwise, the reason we have these is so that we don't get as many false positives.

src/store/model/amazon-de.ts
```diff
@@ -23,34 +24,34 @@ export const AmazonDe: Store = {
 			url: 'https://www.amazon.de/dp/B083JX52VG/'
 		},
 		{
-			brand: 'pny',
-			model: 'xlr8',
+			brand: 'asus',
```
jef, line 27: I find it interesting that you moved a lot of these in lexicographical order, but then some not. Am I reading this right?

jef (changes requested):
Wow, quite a bit of store addon here! This is great. Thanks for doing this.

Overall, looks pretty good. Left some concerns.

Also, what exactly is Twitch outputting?

src/store/model/alternate.ts
PaddseL: as i mentioned in the testing section of this pr's comment, i am happier with the outcome when only checking for `outOfStock`.
german web store barely get any stock atm compared to the na/ca market and if they get any, their status is more like a "available in x to y days" which makes it impossible to search for in the `inStock` label as we don't have regex search yet and most of the shops don't change their "Add to cart" button with different status(es?); e.g. notebooksbilliger (de) always have a "Jetzt vorbestellen" button which translates to "preorder now".

maybe i don't get the point of your argument here, but for my use case this implementation worked like a charm and i got notifications if some stores expected stock and accepted preorders.

src/store/model/amazon-de.ts
PaddseL: yes i started to normalize and sort them somewhere in the middle of my contributions and forgot about them, sorry.

PaddseL: i actually don't know what i tried here first, no shop except one was in correct order.

PaddseL:
> Wow, quite a bit of store addon here! This is great. Thanks for doing this.
> 
> Overall, looks pretty good. Left some concerns.
> 
> Also, what exactly is Twitch outputting?

first of all sorry again for this huge mess of a pr and thanks for your feedback.
i updated most aspects according to your comments.

i don't get your last question, sorry, but i try to explain a bit here.
i mainly implemented the `ChatClient` according to this [example](https://d-fischer.github.io/twitch-chat-client/docs/examples/basic-bot.html), unfortunately you have to do a full oauth2 cycle by hand if you do it like this.

first i initialize `tokenData` with `.env` config values. if a `twitch.json` file exists, i try to get `accessToken`, `expiryTimestamp` and `refreshToken` from it, as they maybe newer. new authentication data gets written to `twitch.js` in the `onRefresh` function, which gets called when `RefreshableAuthProvider` gets new authentication data.

finally, i connect to the twitch chat via `chatClient.connect()`, which returns a promise i neither cannot wait for because not in an async context, nor i have something to do when finished so i tell the linter to ignore it with `void`.

in action it looks like this:
image

in case this did not answer your question, could you specify it a bit more and do you have any further questions?

fatmoo5:
Test your repo and tried out twitch and says sent but nothing is in chat
https://i.imgur.com/J12ZFBC.png

PaddseL:
> Tested your repo and tried out twitch and says sent but nothing is in chat
> https://i.imgur.com/FiW6PnH.png

are you sure your config values are set correctly? `TWITCH_CHANNEL` must be your twitch account name in lowercase, e.g. my twitch name is "Paddsel" and i set it to "paddsel".

fatmoo5:
> > Tested your repo and tried out twitch and says sent but nothing is in chat
> > https://i.imgur.com/FiW6PnH.png
> 
> are you sure your config values are set correctly? `TWITCH_CHANNEL` must be your twitch account name in lowercase, e.g. my twitch name is "Paddsel" and i set it to "paddsel".
Okay so I ran npm run test:notification, it didnt send anything......But then I tried test:series and it worked fine. sorry about that!

PaddseL:
> > > Tested your repo and tried out twitch and says sent but nothing is in chat
> > > https://i.imgur.com/FiW6PnH.png
> > 
> > 
> > are you sure your config values are set correctly? `TWITCH_CHANNEL` must be your twitch account name in lowercase, e.g. my twitch name is "Paddsel" and i set it to "paddsel".
> > Okay so I ran npm run test:notification, it didnt send anything......But then I tried test:series and it worked fine. sorry about that!

No problem, this actually is an issue and i know the cause:
in production, there is a good amount of time between `chatClient.connect()` and the first notification event. when running `npm run test:notification` however, the `chatClient.say()` immedietaly gets called after the `chatClient.connect()`, where the connection is not established yet.

any ideas on a clean solution to fix this?

PaddseL:
i think i came up with a good workaround:
using events now to send messages to twitch chat, `npm run test:notification` works now like production und it is overall a more cleaner solution.

no need to wrap the instantiation of `chatClient` in a check now as it only matters when executing `chatClient.connect()`.

Please let me know what you think :)

johnconnor2020:
THIS IS AWESOME

src/notification/telegram.ts
```diff
@@ -15,13 +15,18 @@ export function sendTelegramMessage(link: Link, store: Store) {
...
+			const results = [];
 
-			try {
-				await client.sendMessage(telegram.chatId, `${Print.inStock(link, store)}\n${givenUrl}`);
-				logger.info('✔ telegram message sent');
-			} catch (error) {
-				logger.error('✖ couldn\'t send telegram message', error);
+			for (const chatId of telegram.chatId) {
+				try {
+					results.push(client.sendMessage(chatId, `${Print.inStock(link, store)}\n${givenUrl}`));
```
jef, line 22: we should make this `async` as well. But that's OK, we can do that in another PR.

src/store/model/alternate.ts
jef: So what happens here sometimes is that they change their out of stock labels and we were getting a lot of false positives at first. We then inverted the logic to only search for add to cart, but that also had false positives. Combining the two ensured that we got a proper alert.

> which makes it impossible to search for in the inStock label as we don't have regex search yet

Usually can be found using the `test:brand`. They usually don't change their classes based on an item.

Ref: https://github.com/jef/nvidia-snatcher/pull/141

jef (dismissed):
HOLY smokes that's a lot of products... Thanks for the great add. I have a few comments there, but that shouldn't hold back this PR any futher.

Thanks for implementing this!

I don't know all of the use cases for Twitch, but I also don't use it that often, so if it makes sense, then great!

jef:
If you're able to resolve these conflicts, we can get this in!

Thank you!

johnconnor2020:
There seems to be an error when checking the zotac card on MediaMarkt and Saturn store:

` warn :: ✖ [mediamarkt] [zotac (3080)] amp extreme holo :: STATUS CODE ERROR 404`

src/notification/telegram.ts
PaddseL: line 16 - 30 is already wrapped in an `async`, do you mean the body of the for loop should be `async` aswell?

src/store/model/alternate.ts
PaddseL: i see, updated the `inStock` label to be more accurate

PaddseL:
I saw a streamer on twitch who is streaming this project for the na/ca market, i thought it would be good idea to do this for the german market as well.
There are many people who want a rtx 30 series card, but don't have enough technical knowledge to run this project, this way they also have a chance of getting one. :) 
It's much easier for the viewers to click on a link in the chat instead of copying the link char by char from the stream and someone in my chat requested this, so i implemented it.
Now, people can stream this for various other markets/regions! :)

johnconnor2020 (approved):

johnconnor2020:
> > I don't know all of the use cases for Twitch, but I also don't use it that often, so if it makes sense, then great!
> 
> I saw a streamer on twitch who is streaming this project for the na/ca market, i thought it would be good idea to do this for the german market as well.
> There are many people who want a rtx 30 series card, but don't have enough technical knowledge to run this project, this way they also have a chance of getting one. :)
> It's much easier for the viewers to click on a link in the chat instead of copying the link char by char from the stream and someone in my chat requested this, so i implemented it.
> Now, people can stream this for various other markets/regions! :)

that sounds great. is he streaming 24/7? do you have a link?

PaddseL:
> that sounds great. is he streaming 24/7? do you have a link?

as far as i know, 24/7

stream for na/ca: [https://www.twitch.tv/falcodrin](https://www.twitch.tv/falcodrin)
my stream for germany: [https://www.twitch.tv/PaddseL](https://www.twitch.tv/PaddseL)

johnconnor2020:
asus store gives me error, i think they changed some stuff on their shop.

src/notification/telegram.ts
jef: Oh, I mean that we should wrap with `Promise` or something.

Kind of like this:

https://github.com/jef/nvidia-snatcher/blob/8098a310925edb5c8738b00259856a2d64e2fec7/src/notification/discord.ts#L30-L35

But I don't even know if `sendMessage` returns a `Promise`, so I really don't know if this would work.

jef (approved):

jef:
Thank you very much!

PR #534, opened 2020-10-17
feat: add incognito mode

GRINSPANKER13:
### Description
Added the ability to use Incognito tabs.
At least with chrome, it uses a cache for sites, incognito forces to reload everything.
This might help better detect when the "add to cart" button is updated.

### Testing
Tested with a normal link with "headless:false" to see if it actually opens in "incognito" and it does.

### DIFF
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -35,7 +35,8 @@ async function lookup(browser: Browser, store: Store) {
 			continue;
 		}
 
-		const page = await browser.newPage();
+		const context = await browser.createIncognitoBrowserContext();
+		const page = (config.browser.isIncognito ? await context.newPage() : await browser.newPage());
 		page.setDefaultNavigationTimeout(config.page.timeout);
 		await page.setUserAgent(config.page.userAgent);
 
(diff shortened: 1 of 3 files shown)


src/store/lookup.ts
```diff
@@ -35,7 +35,8 @@ async function lookup(browser: Browser, store: Store) {
 			continue;
 		}
 
-		const page = await browser.newPage();
+		const context = await browser.createIncognitoBrowserContext();
+		const page = (config.browser.isIncognito ? await context.newPage() : await browser.newPage());
```
jef, line 39: I'd wrap this context within the if statement as well, so we don't create the incognito browser context without needing to.

jef (changes requested):
Looks good overall! Just the one comment. Let me know what you think.

Thanks!

GRINSPANKER13:
> Looks good overall! Just the one comment. Let me know what you think.
> 
> Thanks!

I'm not sure how you would wrap it in an if statement, but every browser has a BrowserContext by default, so I changed the code to assign context to be either default or incognito. Also, closing the page, closed the default BrowserContext, but you have to close the incognito BrowserContext separately. What do you think?

jef:
Ah, that makes a lot of sense to me now. I remember being a little confused when I was playing with the context and passing it around in a feature branch I was working in.

Thanks for the explanation!

jef (approved):

PR #547, opened 2020-10-19
feat: add support for multiple browser user agents

SesioN:
### Description
Added the ability to define a pool of many browser user agents in the environment file.
On each request a random user agent will get used from the defined pool.

The idea is to keep the bot more anonymized and harder to detect by having many user agents instead of just one.

It would be better to have each user agent separated by a line break, but sadly the version of dotenv doesn't support it yet?

Supported in:
https://github.com/bkeepers/dotenv#multi-line-values

But not in:
https://github.com/motdotla/dotenv

Example source for common user agents:
https://developers.whatismybrowser.com/useragents/explore/

**NOTICE**
Breaking change duo to rename from USER_AGENT to USER_AGENTS in env file!
Maybe this feature could create an additional entry instead of rename or just use old name.


jef:
> Breaking change duo to rename from USER_AGENT to USER_AGENTS in env file!
> Maybe this feature could create an additional entry instead of rename or just use old name.

Perhaps just revert to `USER_AGENT`. I don't mind having this come in. Other variables are also non-plural.

jef (changes requested):
Revert `USER_AGENTS` to `USER_AGENT`

jef (approved):
Awesome! Thanks so much for the work! Cool idea.

PR #550, opened 2020-10-20
fix: update gigabyte model `vision oc`

KazanK:
### Description

Purely an informational change for anyone looking to filter by Gigabyte's Vision OC card. Following the readme's filter guide for this model results in a filter for a card that doesn't exist.

### Testing

Running this while filtering by `gigabyte` brand results in searches for `vision oc` cards (along with other Gigabyte cards, as expected), but filtering by the `vision` model like the readme says doesn't result in any searches. Filtering by the `vision oc` model works as expected.


jef:
Ah, makes sense. Thanks for updating!

jef (approved):

PR #558, opened 2020-10-20
feat(store): Add PCComponentes and Amazon-ES Stores.

alvarobasi:
### Description

Added spanish PCComponentes and Amazon-ES stores to the store list along with their product links and stock status states.


jef (approved):
Looks good! Thanks.

PR #563, opened 2020-10-21
fix(store): pccomponents model normalization

### LINKED ISSUE #560: Bug in PCComponentes store
### Description

When selecting `pccomponentes` as one of the stores to track, it doesn't appear in the logs.

I'm using the dockerized application in a Windows 10 machine.

### Possible solution

In the [pccomponentes.ts](https://github.com/jef/nvidia-snatcher/blob/main/src/store/model/pccomponentes.ts) file I chose the following labels for checking status:
```
labels: {
		inStock: {
			container: '#btnsWishAddBuy',
			text: ['Comprar']
		},
		outOfStock: {
			container: '#btnsWishAddBuy',
			text: ['Avísame']
		}
	},
```

However, testing this container selection in the Chrome Console gives the following output:
```
console.log($('#btnsWishAddBuy').text());

$Añadir a lista}¥Añadir al carrito
Comprar
]
```

How does the program filter using the `text: ['Comprar']` property? Is there a contains function or something similar? Maybe the breaklines are hurting the algorithm? Or may the error be in the `text: ['Avísame']` property with the accent in the i? In that case, leaving the `inStock` property alone would be the solution.

alvarobasi:
### Description

Fixes #560 
Fixed bug for PCComponentes shop not matching the established labels.

### DIFF
--- a/src/store/model/pccomponentes.ts
+++ b/src/store/model/pccomponentes.ts
@@ -32,7 +32,7 @@ export const PCComponentes: Store = {
 		},
 		{
 			brand: 'asus',
-			model: 'tuf gaming',
+			model: 'tuf',
 			series: '3080',
 			url: 'https://www.pccomponentes.com/asus-tuf-geforce-rtx-3080-10gb-gddr6x'
 		},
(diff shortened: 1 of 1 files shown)


jef (approved):
Thanks for updating an fixing! I should make these types...

PR #573, opened 2020-10-22
fix: asus store links

fatmoo5:
Oddly enough fixes the asus 503 error


fatmoo5:
Maybe not accept my pull but remove the / in the asus store and it fixes it.

jef (approved):
I've seen problems with web servers not accepting ending forward slashes. Thanks for looking into this.

PR #595, opened 2020-10-24
feat: allow filtering per model by a specific series

SesioN:
### Description
Possibility to filter additionally a model by a series.

**Use-Case:**
Searching for 3 cards: the founders edition 3090, asus strix oc 3080 and 3090.
Both models are generally available for both series. (e.g founders edition 3080, founders edition 3090 etc.)

**Example Config for Use-Case:**
STORES="asus,nvidia"
SHOW_ONLY_BRANDS="asus, nvidia"
SHOW_ONLY_MODELS="rog strix oc, founders edition:3090"

Result:
nvidia-snatcher searches for: rog strix oc 3080, rog strix oc 3090, founders edition 3090

**Limitations**
Additional filtering only works for series that are globally enabled and not filtered out by SHOW_ONLY_SERIES.


jef:
> Additional filtering only works for series that are globally enabled and not filtered out by `SHOW_ONLY_SERIES`

Makes sense. I imagine someone would use one or the other.

jef (approved):
Thank you! This is pretty nice!

PR #605, opened 2020-10-26
feat(store): support for coolmod

alvarobasi:
Added Coolmod shop to the retailers list. Only 3080. 3070s will be uploaded when they come up.


jef (approved):
Looks good! Thank you!

PR #608, opened 2020-10-26
docs: add origin country to non-US stores

designgears:
### Description

add origin country to non-US stores in the readme.


jef (approved):
Awesome. Thanks for this clarification!

Issue #600, opened 2020-10-25
"STORES" environment variable doesnt work at start


TheTabKey:
Now goes through every store at the beginning, then after initial spam it will go back to the configured variables.

Macaddict42:
Is this really fixed? I'm still experiencing this.

jef:
Not sure, seems to work fine for me. I would make sure you have the latest code.

Macaddict42:
![ns_log](https://user-images.githubusercontent.com/38168730/97332870-2e5cc900-1838-11eb-9d19-eebec2a67b64.PNG)
This is what it looks like for me. I'm on the latest source and have run npm install after a git pull. Is there a cache I should try clearing?

jef:
Interesting. OK, what does your `.env` look like?

Macaddict42:
# ** All configuration variables are optional **
# Read https://github.com/jef/nvidia-snatcher#customization for help on customizing this file
#############################################################################################

BROWSER_TRUSTED="false"
COUNTRY="usa"
DESKTOP_NOTIFICATIONS="false"
EMAIL_PASSWORD="redacted"
EMAIL_TO="redacted"
EMAIL_USERNAME="redacted"
HEADLESS="true"
IN_STOCK_WAIT_TIME="0"
LOG_LEVEL="info"
LOW_BANDWIDTH="false"
MAX_PRICE_SERIES_3070="600"
MAX_PRICE_SERIES_3080="900"
MAX_PRICE_SERIES_3090="1600"
MICROCENTER_LOCATION="web"
NVIDIA_ADD_TO_CART_ATTEMPTS="10"
NVIDIA_SESSION_TTL="60000"
OPEN_BROWSER="true"
PAGE_BACKOFF_MIN="10000"
PAGE_BACKOFF_MAX="3600000"
PAGE_SLEEP_MIN="5000"
PAGE_SLEEP_MAX="10000"
PAGE_TIMEOUT="30000"
PHONE_CARRIER="tmobile"
PHONE_NUMBER="5035503453"
PLAY_SOUND="alert.mp3"
PUSHOVER_TOKEN="redacted"
PUSHOVER_USER="redacted"
PUSHOVER_PRIORITY="1"
SCREENSHOT="true"
SHOW_ONLY_BRANDS="asus, nvidia"
SHOW_ONLY_MODELS="rog strix, rog strix oc, strix, tuf, tuf oc, founders edition"
SHOW_ONLY_SERIES="3080"
[...]

PR #614, opened 2020-10-27
refactor: easier comparables, fix audits

### LINKED ISSUE #620: Stores showing up that I'm not searching for
Im running a snatcher on a VM and I keep getting these messages. It says no store links found but i only have  bestbuy, amazon, newegg, newegg-ca, bestbuy-ca, asus.

**[5:03:27 PM] info :: ℹ selected stores: bestbuy, amazon, newegg, newegg-ca, bestbuy-ca, asus
[5:03:27 PM] info :: ℹ selected models: ftw3, ftw3 ultra, xc3 black, xc3, xc3 ultra, tuf, tuf oc, ventus 3x, ventus 3x oc, founders edition
[5:03:27 PM] info :: ℹ selected series: 3080
[5:03:28 PM] warn :: ✖ in order to receive sms alerts, email notifications must also be configured
[5:03:35 PM] warn :: ✖ [amazon-uk] [setup (3080)] :: NO STORE LINKS FOUND
[5:03:35 PM] warn :: ✖ [amazon-uk] [setup (3080)] :: NO STORE LINKS FOUND
[5:03:37 PM] warn :: ✖ [novatech] [setup (3080)] :: NO STORE LINKS FOUND
[5:03:37 PM] info :: ✖ [bestbuy-ca] [msi (3080)] ventus 3x :: OUT OF STOCK
[5:03:38 PM] warn :: ✖ [box] [setup (3080)] :: NO STORE LINKS FOUND
[5:03:39 PM] info :: ✖ [newegg] [asus (3080)] tuf :: OUT OF STOCK
[5:03:40 PM] info :: ✖ [aria] [setup (3080)] :: FOUND 4 STORE LINKS
[5:03:41 PM] info :: ✖ [bestbuy-ca] [evga (3080)] xc3 ultra :: OUT OF STOCK**

jef:
### Description

- Fixes #620
- Related to #600
	- This will go in a later refactor
	- For now, I am putting the link builder in debug mode
- Bumps node version to 15.0.1
- Removes `rimraf` from `build` task
- `npm audit` now has 0 vulnerbilities
- Fixes issue with Web UI not saving settings


Doridian:
This PR will work, however from what I can see, if you were to start with a certain store disabled and then enable it in WebUI, the store would not have any links defined.

Also if you do this, you should do
`stores.set(storeData.name, store);`
and put it inside the `if (store) {` to avoid putting blank/undefined stores into the `stores` map.

jef:
> the store would not have any links defined.

I understand what you're saying, but I don't see how. Could you elaborate on this?

Doridian:
Sure. I haven't 100% validated this, but basically here is what I'm thinking.

The change https://github.com/jef/nvidia-snatcher/pull/614/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80L58-R58
pretty much  makes it only loop over enabled stores when it runs that codepath (as `getStores()` only contains stores defined in the config, and thus enabled).
So if a store is not enabled while snatcher starts up, that codepath won't run ever.
This means, there will be no loop started for the store, and the startupAction won't run for the store.

jef:
I see in your code now within the `html` file. It's directly utilizing the `config.store.stores` as we talked about. Okay, in that case, perhaps I do similar function call outside of the config and use it there. Or write up an observer. Either way I suppose would work.

I think overall, some of this stuff could use some extra love, but I really appreciate your feedback!

jef:
Okay, I updated the `handleAPI()` function to `updateStores()` when the config gets updated. This should resolve that issue.

Granted, this solution isn't ideal and there is a better way to do this entirely, but this should at least bandage this up until we refactor some of this config.

jef:
Okay, I see your point as well. We never go over that loop of stores again, regardless of where it's coming from.

I'm going to update the place we need to, along with some other things. This could _really_ use some TLC...

jef:
Gonna merge this in later if there aren't any comments! Thanks for the help @Doridian.

PR #627, opened 2020-10-29
chore: add 3070 cards and correct models on 3080

KillahInstinct1:
There was a few errors in the models on the 3080-series, and I added some of the cards that I could from:

https://azerty.nl/nvidia-rtx-3070-launch


jef (approved):
Awesome! Thanks!

PR #634, opened 2020-10-29
chore(bestbuy-ca): add 3070 founders edition

morinted:
### Description

Founders Edition is only sold through Best Buy in Canada. Link was only added this morning at release time so it wasn't included in the previous Canadian 3070 PR.

+ Add Best Buy Canada and Newegg Canada to read me.

### Testing

Used to watch the page after the card went out of stock.


jef (approved):
Thank you!

PR #637, opened 2020-10-29
fix: update pushbullet import to match package

zinefer:


jef:
Awesome! Thank you for fixing this. That was dumb on my part.

jef (approved):

PR #647, opened 2020-10-30
Newegg disabled add to cart urls for 30 series

TheTabKey:
Newegg disabled add to cart urls for 30 series

### DIFF
--- a/src/store/model/newegg.ts
+++ b/src/store/model/newegg.ts
@@ -24,336 +24,336 @@ export const Newegg: Store = {
 		},
 		{
 			brand: 'asus',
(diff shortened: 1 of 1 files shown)


jef:
Do you mind elaborating a little on why we want to do this?

TheTabKey:
Newegg no longer has the add to cart url for 3080 working. So like this: 
https://secure.newegg.com/Shopping/AddtoCart.aspx?Submit=ADD&ItemList=N82E16814126452
Should say sorry no longer in stock and ask you to remove the item.
I will update with a link to show correct behavior of out of stock item when I can find one.

v1ct0rv:
Would be great if we received in the message the cart url and the producto url

PR #659, opened 2020-11-01
chore(proshop-de): add 3070 cards

Tomson124:
### Description
Adding the 3070 cards from Proshop Germany (prohsop-de)

### Testing

Navigate to Proshop.de page to confirm URL hasn't changed or product has disappeared


jef (approved):
Woo hoo!!! Thanks for adding! :rocket:

PR #665, opened 2020-11-02
chore: add more 3070 cards to media markt and notebooksbilliger, fix false positives on media markt

Tomson124:
### Description

Adding more 3070 cards to media markt and notebooksbilliger
Fixing false positives due to media markt often not label availability the same on all cards


jef (approved):
Awesome! Thanks for the additions!

PR #664, opened 2020-11-02
feat: newegg realtime api

jbwilliams1:
### Description

Added realtime inventory lookup for Newegg cards

Resolves #646 
Resolves #643 

### Testing

Used an in-stock card for the test series, validated it worked on an in-stock and out of stock card. Runs normally on out-of-stock cards

### New dependencies

None (well, node-fetch, but it's bundled in)

PR #674, opened 2020-11-03
feat(newegg-ca): realtime api

kdelalic:
### Description

Added real time inventory lookup for Newegg Canada cards
Based on #664 
Also added 5 new cards

### Testing

Runs without issue when card is out of stock
Same api as newegg.com


jef (approved):
Awesome! Thanks for contributing :)

PR #675, opened 2020-11-03
feat(asus): realtime api

### LINKED ISSUE #633: App reporting cards in stock at Asus store when they are not
### Description

App is reporting Asus ROG Strix and TUF 3080 cards in stock, but when I visit the site, they are not in stock. Note there may be other series that also show up as in stock, but I'm only checking for 3080's.

jbwilliams1:
### Description

Fixes #633 

Derived from Newegg API realtime lookup code structure 

### Testing

Tested against in-stock 5700xt and out-of-stock 3070. Does not interfere with normal operation.

### New dependencies

None

### DIFF
--- a/src/store/model/asus.ts
+++ b/src/store/model/asus.ts
@@ -1,4 +1,5 @@
 import {Store} from './store';
+import fetch from 'node-fetch';
 
 export const Asus: Store = {
 	labels: {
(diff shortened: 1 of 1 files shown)


jef (approved):
These PRs are really awesome. I really appreciate you taking the time and catching all the XHR requests.

drafton:
Is it possible to apply the same changes to the store `asus-de`? Or is it not supported there?

jbwilliams1:
@drafton doesn't look like their DE store has this type of API

drafton:
Too bad, thanks for checking anyway

PR #680, opened 2020-11-03
fix(filterBrandsSeriesModels): always loop all stores

Doridian:
### Description

This should always have all possible combinations. And also set min/max page sleep to sane defaults.

### DIFF
--- a/src/store/model/index.ts
+++ b/src/store/model/index.ts
@@ -104,12 +104,12 @@ const models = new Set();
 const series = new Set();
 const stores = new Map();
 
-function filterBrandsSeriesModels(stores: Map<string, Store>) {
+function filterBrandsSeriesModels() {
 	brands.clear();
 	series.clear();
 	models.clear();
 
-	for (const store of stores.values()) {
+	for (const store of storeList.values()) {
 		for (const link of store.links) {
 			brands.add(link.brand);
 			series.add(link.series);
@@ -175,7 +175,7 @@ export function updateStores() {
 		}
 	}
 
-	filterBrandsSeriesModels(stores);
+	filterBrandsSeriesModels();
 	printConfig();
 }
 


jef:
Can you explain this a little more to me? I'd figure we wouldn't want to do extra work that isn't needed.

Doridian:
It doesn't really do any more work. Unless you mean adding some strings to `Set`s (which is just done once on startup, and makes the WebUI list all cards, regardless of which stores you picked, which is correct).
Basically, the issue is without this in place, some stores have `undefined` as their minSleep/maxSleep. Which, due to the `setTimeout` used, can lead to an absolute mess in terms of CPU load. (`undefined` or `NaN` in `setTimeout` basically means "re-run immediately").

Doridian:
Essentially, what I'm saying is that without at least
```
		if (store.minPageSleep === undefined) {
			store.minPageSleep = defaultStoreData.minPageSleep;
		}

		if (store.maxPageSleep === undefined) {
			store.maxPageSleep = defaultStoreData.maxPageSleep;
		}
```
being run on every store (no matter if enabled or not), if you don't enable all stores, the process starts consuming way more CPU than it should just ramming infinite loops.

jef:
Oh man, this really needs a refactoring... I we shouldn't even be evaluating these stores. They should be options in the UI, but not evaluated.

Such a shame. Thanks for the fix Doridian.

jef (approved):

heypiotr:
FWIW, I "fixed" this in a different way in my local version:

```diff
diff --git a/src/index.ts b/src/index.ts
index 6c377fa..0a12cdc 100644
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,7 +7,7 @@ import {logger} from './logger';
 import puppeteer from 'puppeteer-extra';
 import resourceBlock from 'puppeteer-extra-plugin-block-resources';
 import stealthPlugin from 'puppeteer-extra-plugin-stealth';
-import {storeList} from './store/model';
+import {getStores} from './store/model';
 import {tryLookupAndLoop} from './store';
 
 puppeteer.use(stealthPlugin());
@@ -55,7 +55,7 @@ async function main() {
                headless: config.browser.isHeadless
        });
 
-       for (const store of storeList.values()) {
+       for (const store of getStores().values()) {
                logger.debug('store links', {meta: {links: store.links}});
                if (store.setupAction !== undefined) {
                        store.setupAction(browser);
```

`storeList` is all the stores, `getStores()` is the filtered list.

jef:
@heypiotr, just seeing this now. Thanks for the recommendation!

PR #696, opened 2020-11-05
feat(amd): add env vars and series filtering

njgreb:
- update .env example
- add MAX_PRICE_SERIES docs for AMD cpu's
- add ryzen cpu's as defaults if SHOW_ONLY_SERIES is empty

### Description

Adds env vars for price max on all AMD Ryzen Zen 3 cpus supported
Adds Zen 3 cpu's to the set of defaults if not filtered by config.


jef:
I think we'll also need to update this too: https://github.com/jef/streetmerchant/blob/5b3e95acdf2f427a6098f15b9fbe65e88354a678/src/store/lookup.ts#L161-L185

But I'll make a following push! Thanks for adding this in! :D

jef (approved):

PR #698, opened 2020-11-06
feat(store): add zen 3 to more stores

serg06:
### Description

Changes:
- Added amazon.ca
- Added newegg.ca
- Noticed an incorrect item ID on newegg.com so fixed that too

Motivation: To get myself a 5950x! Let's go!!

### Testing

Double-checked URLs, double-checked product numbers, and started running it. Seems to run fine.


jef (approved):
Awesome! Thanks @serg06!

PR #702, opened 2020-11-06
feat(store): add zen 3 to german stores

hiivin:
## Description

I added the zen 3 processors to some german stores: alternate, amazon-de, caseking, computeruniverse, cyberport, mindfactory (only 5600x and 5800x, I could not finde the other links), notebooksbilliger, proshop-de .

### Testing

I let the streetmerchant check the stock for the zen 3 processors on these stores and it worked fine.

Screenshot 1

Screenshot 2


jef (approved):
Awesome! Thanks!

PR #709, opened 2020-11-06
feat(ps5): add console and digital edition

mven:
### Description

- Add PS5 Console and PS5 Digital Edition support
- Add Walmart, Target stores


jef:
Sure! Sounds good to me. This list is quickly growing...

jef (approved):

happytambi:
It would be amazing if you could get the xbox support live before launch day on the 10th.  I'm super new to all of this, If someone can point me in the right direction Id be glad to give it a shot my self.  Thanks in advnace!

PR #715, opened 2020-11-07
feat(store): add ps5 to german stores

hiivin:
### Description

I added the PS5 to the German stores alternate, amazon-de, mediamarkt and saturn.

### Testing

Seemed to work fine:

Screenshot 1


jef (approved):
Looks good! Thanks!

Issue #694, opened 2020-11-05
Regression: Generate links for all stores


jef:
### Description

Currently, when `streetmerchant` starts up, links are generated for various stores. This behavior shouldn't happen if the user doesn't use those stores.

Refs:

- https://github.com/jef/streetmerchant/issues/683
- https://github.com/jef/streetmerchant/issues/691#issuecomment-722660867

### Possible solution

Make sure only links are generated for stores selected.

garza-utsa:
@jef I was able to get the develop branch searching again by disabling the amazon-uk store, there only seems to be three link entries in there (one test and two ps5 sony), it was causing a hang after displaying "Failed to parse card: PNY GeForce ..." error line.

Is there a reason why links are generated for stores not even selected in the first place (I'm in the US with only 'newegg,amazon,bestbuy' set in my stores.

garza:
@jef I was able to get the develop branch searching again by disabling the amazon-uk store, there only seems to be three link entries in there (one test and two ps5 sony), it was causing a hang after displaying "Failed to parse card: PNY GeForce ..." error line.

jef:
Yeah, this is part of the regression of building links. I haven't spent anytime working on it, but should be fixed in the near future! Thanks for the comment, it will be helpful.

taefighter:
Is there a workaround available for this? I have not been able to use streetmerchant for a few days because of this issue. I just get a 
"failed to parse..." error and the application stops.

garza:
@taefighter do a git pull on main, i haven't had a problem since Friday with the latest code

jef:
> Is there a workaround available for this? I have not been able to use streetmerchant for a few days because of this issue. I just get a
> "failed to parse..." error and the application stops.

This prohibit from the app from running at all. It just slows down and shows errors at startup.

taefighter:
> @taefighter do a git pull on main, i haven't had a problem since Friday with the latest code

Thank you for letting me know. I am still getting the parsing error at startup. However, what @jef said tells me I wasn't giving the app enough time to process. I am leaving it running for a while to see if it will work past the errors at the start. Thank you!

PR #719, opened 2020-11-07
feat(amazon): add more 3070s

vp-cshow:
### Description

Adds a few 3070s on amazon-us

NEW: 
Model for pny: xlr8 uprising
Model for asus: dual (there's two pages for this model at the moment, I went with the less sketchy looking one) 

### Testing

Only tested OOS logic, couldn't verify in-stock for obvious reasons

### Issues

~~I was running into some error logs for a PNY card, but it appears to be related to the e-buyer automatic link generation, who's logic doesn't apply to amazon cards~~ Unrelated issue per #694

### DIFF
--- a/src/store/model/amazon.ts
+++ b/src/store/model/amazon.ts
@@ -17,10 +17,59 @@ export const Amazon: Store = {
 	links: [
 		{
 			brand: 'test:brand',
-			cartUrl: 'https://www.amazon.com/gp/aws/cart/add.html?ASIN.1=B07TDN1SC5&Quantity.1=1',
+			cartUrl: 'https://www.amazon.com/gp/aws/cart/add.html?ASIN.1=B07TDN1SC5&Quantity.1=1:',
 			model: 'test:model',
 			series: 'test:series',
 			url: 'https://www.amazon.com/dp/B07TDN1SC5'
+		},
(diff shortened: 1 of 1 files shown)


src/store/model/amazon.ts
```diff
@@ -22,6 +22,55 @@ export const Amazon: Store = {
...
+			series: '3070',
+			url: 'https://www.amazon.com/dp/B08HBJB7YD'
+		},
+		{
+			brand: 'pny',
+			cartUrl: 'https://www.amazon.com/gp/aws/cart/add.html?ASIN.1=B08HBJB7YD&Quantity.1=1',
+			model: 'xlr8 uprising',
+			series: '3070',
+			url: 'https://www.amazon.com/dp/B08HBJB7YD'
+		},
```
jef, line 59:
```suggestion
```

Duplicate card from above

jef (changes requested):

src/store/model/amazon.ts
vp-cshow: Resolved, those two cards are on the same product page under different "styles". 🤦‍♂️

jef (approved):
Awesome! Thank you very much. Looks great!

PR #721, opened 2020-11-08
feat(store): add `arlt` and `amd-de`

hiivin:
### Description

I have added the store Arlt with some 3070s, 3080s, 3090s and the zen 3 processors. I have also added the german AMD as a store with the zen 3 processors.

### Testing

I tested the status check for Arlt and AMD-de and I tested the price recognition for Arlt with 3 and 4 digit numbers.

screenshot 1

screenshot 2

screenshot 3


jef (approved):
Awesome! Thank you very much!

PR #729, opened 2020-11-09
feat(galaxus): add zen 3

hiivin:
### Description

I fixed a typo in the Arlt store (5950x was labelled as 5900x) and I added the Galaxus (DE) store with the zen 3 links.

### Testing

I tested items that are in stock and items that are not in stock, as well as price recognition for 3-digit and 4-digit amounts.


jef (approved):
Thanks for making the fix and adding to the list!

PR #737, opened 2020-11-10
feat(target): add additional selector for 'ship it'

mven:
### Description

Add `Ship it` button selector to Target store


jef (approved):
Great! Thanks!

Issue #273, opened 2020-09-24
Possible to add function for two phon numbers?


jordanhgaytan:
Have two phones, one personal and one work phone. Would be awesome to have both be able to receive a notification!

Is this remotely possible?

jordanhgaytan:
spelling

geman220:
It's possible but it's going to be an extremely low priority in comparison to other enhancements unless someone wants to fork and add it.

adolfox:
@jef Is this still something you want to implement?  I was thinking of taking a crack at it.

adolfox:
I'm doing Hacktoberfest.  I need two more PRs :)

jef:
> I'm doing Hacktoberfest. I need two more PRs :)

Go for it! I don't mind having this in here. Makes sense to me.

adolfox:
So I was thinking that in order to keep the .env compatibility intact that I'd just extend on the current variables in the .env.  I was thinking that people that want multiple numbers, would just comma separate the list of number and then in the corresponding order, comma separate the carriers, something like:

PHONE_NUMBER="5121234567, 9876543987"
PHONE_CARRIER="tmobile,att"

Does that work for you?

jef:
Yep, I think that's great. No spaces between `,` (probably a typo), but that looks great. Perhaps we can also `trim()` these environment variables in the `envOr*()` functions.

adolfox:
Yeap, the space was a mistake.  Sounds good.  I'll get started on it.  And yeah, I'll trim() in case anyone includes spaces like I just did.

PR #446, opened 2020-10-04
feat: add ability to send SMS to multiple phone numbers with same or different carriers

adolfox:
### Description

Resolves #273 

You can now list multiple phone number in .env.  If only one carrier is listed, it'll use that one carrier for all phone numbers.  If more than one carrier is listed, the list of carriers has to be in the same order as the phone numbers listed.

### Testing

did a test with one number and one carrier, it passed

did a test with two numbers and one carrier, it passed
two_sms_sent-10-04 075044

did a test with two numbers and two carriers(the carriers were the same, but listed twice), it passed

did a test with three numbers, with spaces after the commans, and one carrier, it passed
test_with_three_numbers_and_spaces

test_with_three_numbers_and_spaces-2

did a test with three numbers, with spaces after the commas, and three carriers, with spaces after the commas, it passed
test_with_three_numbers_and_spaces_and_three_carriers_with_spaces

did a test for the warning if it's not one carrier and the number or carriers doesn't match the number of phone numbers.
warning_if_carrier_is_not_one_or_same_number_or_phone_numbers

did a test of just running the app for a few minutes, no issues.

PR #738, opened 2020-11-10
feat(notification): support for multiple phone numbers

benw25:
### Description

Fixes https://github.com/jef/streetmerchant/issues/273.
Update to stale pull request https://github.com/jef/streetmerchant/pull/446

Multiple carriers & phone numbers are now supported in the .env file.   Number of carriers must strictly match the number of phone numbers (documented in readme)

Error handling was changed a bit in sms.ts:
- if number of carriers does not match number of phones (even if one is 0), logger will warn
- if there is a difference in `carrier.length` or `number.length`, sendSms will say:
> (carrier) is not associated with a number

or
> (number) is not associated with a carrier
- phone.availableCarriers.has(carrier) has been moved to trigger a logging error inside of sendSms, and this logging function has been removed from generateAddress
- any of these above errors will continue and skip the rest of sendSms
- generateAddress now takes in a number and a carrier as arguments
- sendMail error callback now has the specific number & carrier to provide more info as to what happened

Misc changes were added to fix linting errors:
- `max-params: 0` (unlimited) to account for xo linting showing an error in `maxPrice` in logger.ts.  5 is more than the apparent default of 4.
- because templating was added, I removed the backslash in the error logging for 
> couldn't send sms
### Testing

- Tested multiple phone numbers & carriers (0 & 0, 1 & 1, 2 & 2), and mismatched combinations of the two (0 & 1, 1 & 0, ...etc.).
- Tested with carriers that do not match the list of availableCarriers (e.g. "goog")
- Tested with spaces in the envs, seems like this is already covered in the .trim in envOrArray
- All the above was run with `npm run test:notification`
- Ran the app with multiple numbers/carriers with no issues

### DIFF
--- a/src/notification/sms.ts
+++ b/src/notification/sms.ts
@@ -10,42 +10,55 @@ if (config.notifications.phone.number && !config.notifications.email.username) {
 
 const [email, phone] = [config.notifications.email, config.notifications.phone];
 
+if (phone.carrier.length !== phone.number.length) {
+	logger.warn('✖ the number of carriers must match the number of phone numbers');
+}
+
 export function sendSms(link: Link, store: Store) {
-	if (phone.number) {
-		logger.debug('↗ sending sms');
-		const carrier = phone.carrier;
+	for (let i = 0; i < Math.max(phone.number.length, phone.carrier.length); i++) {
+		const currentNumber = phone.number[i];
+		const currentCarrier = phone.carrier[i];
 
-		if (carrier && phone.availableCarriers.has(carrier)) {
-			const mailOptions: Mail.Options = {
-				attachments: link.screenshot ? [
-					{
-						filename: link.screenshot,
-						path: `./${link.screenshot}`
-					}
-				] : undefined,
-				from: email.username,
-				subject: Print.inStock(link, store, false, true),
-				text: link.cartUrl ? link.cartUrl : link.url,
-				to: generateAddress()
-			};
-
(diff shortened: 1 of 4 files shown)


jef:
> * `max-params: 0` (unlimited) to account for xo linting showing an error in `maxPrice` in logger.ts.  5 is more than the apparent default of 4.

Maybe this could be something more reasonable like `10` for now? I'll change that in a later PR. Thanks for updating :)

src/notification/sms.ts
```diff
@@ -10,42 +10,55 @@ if (config.notifications.phone.number && !config.notifications.email.username) {
...
+		const currentNumber = phone.number[i];
+		const currentCarrier = phone.carrier[i];
+
+		if (!currentNumber) {
+			logger.error(`✖ ${currentCarrier} is not associated with a number`);
+			continue;
+		} else if (!currentCarrier) {
+			logger.error(`✖ ${currentNumber} is not associated with a carrier`);
+			continue;
+		}
```
jef, line 28: Since we are doing the check on sizing in `sms.ts:13-15`, I'm not sure if this is needed.

jef (approved):
Great! I appreciate your work on fixing this. I really appreciate the help!

src/notification/sms.ts
jef: Oh, I see now. It's because it still allows the user to have misconfigured phone config.

PR #742, opened 2020-11-10
docs: add zen3 filter to `SHOW_ONLY_SERIES`

BaelfireNightshd:
docs: Added an example of how to add a zen3 cpu to the filter to the readme.

### Description

I noticed that I got the syntax for adding a Ryzen cpu to the filter wrong and thought the readme could use a subtle update to show how to do it. 

### Testing

N/A


jef (approved):
Makes sense to me. Sorry for the confusion :sweat_smile: Thanks!

PR #756, opened 2020-11-11
chore(pccomponentes): add ryzen cpus

Crstian19:


jef (changes requested):
Just run a `npm run lint:fix` and this should do it. Thanks!

jef (approved):
Great! Thank you!

PR #759, opened 2020-11-11
feat: add option to disable auto add to cart

njgreb:
### Description

As sites (like Best Buy) add various queueing methods, auto add to cart might not be wanted by all users.  Let this be disabled by env var.

### Testing

Tested against new egg with a known to be in stock item.


src/store/lookup.ts
```diff
@@ -117,7 +117,10 @@ async function lookupCard(
 	}
 
 	if (await lookupCardInStock(store, page, link)) {
-		const givenUrl = link.cartUrl ? link.cartUrl : link.url;
+		const givenUrl =
+			link.cartUrl && !envOrBoolean(process.env.DISABLE_ADD_TO_CART, false)
```
jef, line 121: I still think we should put this in config and export that rather than the `envOrBoolean`.

jef (changes requested):

src/store/lookup.ts
njgreb: Agreed, I also switch the env var to be a positive rather than a negative so I didn't have to use a ! in the ternary.  Reads cleaner in my book.

jef (approved):
Thank you very much!!! This is great.

PR #763, opened 2020-11-11
fix(walmart): ps5 digital url

jadenv:
### Description
`ps5 console` and `ps5 digital` models both incorrectly shared the same URL in the Walmart model.

This PR corrects `ps5 digital` URL to point to the correct product page.


mven:
FYI - the former link is the one being served on their PS5 promo page: https://www.walmart.com/cp/playstation-5/3475115

Most likely both versions serve the same content.

jef (approved):
Thanks!

PR #765, opened 2020-11-11
add amazon-fr

Zyde22:
Added amazon.fr.


jef:
How come you removed this? Seemed good to me!

PR #767, opened 2020-11-11
feat(store): add vsgamers

Crstian19:


src/store/model/index.ts
```diff
@@ -53,6 +53,7 @@ import {Scan} from './scan';
 import {Store} from './store';
 import {Target} from './target';
 import {Very} from './very';
+import {vsgamers} from './vsgamers';
```
jef, line 56:
```suggestion
import {VsGamers} from './vsgamers';
```

jef (changes requested):
Hi there! Thanks for contributing. I'd rather not maintain a `docker-compose.yaml` file at this time.

Otherwise, looks good! Just need to fix the linting.

jef:
Thank you!

jef (approved):

PR #790, opened 2020-11-13
feat(asus): add backoff codes

### LINKED ISSUE #789: Add backoff values for Asus store
### Description

When including Asus in the list of stores, streetmerchant would successfully check the site 1-3 times, before repeatedly getting `503` status codes. I think it's running up against a rate limiter.

### Possible solution

I'm submitting a PR now that adds the usual backoff codes, plus 503, so that streetmerchant can still successfully check the Asus store. I've tried this and it works great.

judy:
### Description

Fixes #789 

I added 503, plus the usual signals that we're getting rate limited, since I'm getting that constantly after 1-2 successful
requests. Works fine when adding this status code.

### Testing

Add `asus` to the `STORES` env var. Before this PR I was getting 503s after a minute or two, and now it receives a 503, backs off, and scrapes more slowly. No more 503s!

### DIFF
--- a/src/store/model/asus.ts
+++ b/src/store/model/asus.ts
@@ -2,6 +2,7 @@ import {Store} from './store';
 import fetch from 'node-fetch';
 
 export const Asus: Store = {
+	backoffStatusCodes: [403, 429, 503],
 	labels: {
 		inStock: {
 			container: '#item_add_cart',


jef (approved):
Oh awesome. This was definitely needed. Thanks for adding!

PR #806, opened 2020-11-15
feat(store): add `PlayStation` store

### LINKED ISSUE #804: Add Support for Playstation Direct Store for the PS5
### Description

Add Support for the Playstation Direct Store for the PS5 Disc and the PS5 Digital Editions

### Possible solution

Here are the PS5 Product URLs that display 'Out of Stock' currently and the 'Add' Button is not visible:

PS5 Disc Edition:
https://direct.playstation.com/en-us/consoles/console/playstation5-console.3005816

PS5 Digital Edition:
https://direct.playstation.com/en-us/consoles/console/playstation5-digital-edition-console.3005817

Here is a Product where the 'Add' Button is visible:

https://direct.playstation.com/en-us/accessories/accessory/dualsense-wireless-controller.3005715

A Queuing System Workaround will be required to be coded into StreetMerchant to get into the Playstation Direct store to check availability when PS5 are about to become available. As of today, The Playstation Direct Store implements a Queuing System which gets put into place typically 30 minutes to an Hour before both Consoles are about to be available for purchase. In order to get into the Playstation Direct Store to check availability in this situation and get around the Queueing System, One can manually install the 'UBlock' Google Chrome plugin and add below filters to block/ignore all requests involving:

queue
queue-it.net
*queue-it.net

With this workaround, Even when the queuing System is in place, One can get into the Playstation Direct store and look at the Product Page to see if it is available. I don't know how this could be integrated/coded into the StreetMerchant Solution.

J-Hinton:
### Description

- Add PlayStation store for PS5 consoles (closes #804)

I have not added support for the `outOfStock` label, due to Playstation store output both 'Add', and 'Out of Stock' where the appropriate stock label is then hidden on load. `lookupCardInStock` config is set to `requireVisible: true` for the `inStock` label but not any other label, meaning I can only implement an `inStock` label alongside using `realTimeInventoryLookup` for validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ environment variables are **optional**._
 | Office Depot | `officedepot`|
 | Overclockers (UK) | `overclockers`|
 | PCComponentes (ES) | `pccomponentes`|
+| PlayStation (ES) | `playstation`|
 | PNY | `pny`|
 | Proshop (DE) | `proshop-de`|
 | Proshop (DK) | `proshop-dk`|
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ environment variables are **optional**._
 | Office Depot | `officedepot`|
 | Overclockers (UK) | `overclockers`|
 | PCComponentes (ES) | `pccomponentes`|
-| PlayStation (ES) | `playstation`|
+| PlayStation | `playstation`|
 | PNY | `pny`|
 | Proshop (DE) | `proshop-de`|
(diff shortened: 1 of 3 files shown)


jef (approved):
Awesome! Thanks for the implementation. I appreciate the API calls as well.

noblesin:
That was awesomely fast... It works great. Will try to think of ways to implement Out of Stock. Thank you so much!

PR #812, opened 2020-11-15
feat(playstation): add `outOfStock` label

### LINKED ISSUE #809: Playstation store not working
Just doesnt seem to want to search the playstation store, and cant seem to find out why.
fBvtoEu

searches every other store for ps5

edit: looks like playstation.ts needs to be updated

J-Hinton:
### Description

- Added outOfStock label to PlayStation store by using `:not(.hide)` class selector - closes #809

### DIFF
--- a/src/store/model/playstation.ts
+++ b/src/store/model/playstation.ts
@@ -4,8 +4,12 @@ import fetch from 'node-fetch';
 export const PlayStation: Store = {
 	labels: {
 		inStock: {
-			container: '.productHero-info .button-placeholder',
+			container: '.productHero-info .add-to-cart:not(.hide)',
 			text: ['Add']
+		},
+		outOfStock: {
+			container: '.productHero-info .out-stock-wrpr:not(.hide)',
+			text: ['Out of Stock']
 		}
 	},
 	links: [


TheTabKey:
Ive noticed default format is xml:
https://api.direct.playstation.com/commercewebservices/ps-direct-us/products/productList?fields=BASIC&productCodes=3005817

However it can be changed to json as show here:
https://api.direct.playstation.com/commercewebservices/ps-direct-us/products/productList?fields=BASIC&format=json&productCodes=3005817

Does that matter at all?

J-Hinton:
@TheTabKey  During testing, I've only seen the API return as JSON - also to note, the PlayStation store also uses the same headers and expects to receive JSON... Are you experiencing XML being returned? 
If so, I'd suggest we can set the HTTP request header to accept only `application/json`.

TheTabKey:
> @TheTabKey During testing, I've only seen the API return as JSON - also to note, the PlayStation store also uses the same headers (`Accept: */*`) and expects to receive JSON... Are you experiencing XML being returned with header `Accept: */*`?
> If so, I'd suggest we can set the HTTP request header to accept only `application/json`.

It works great, sorry about that.

kyleroath:
Solid work @J-Hinton! Just tested it now, working flawlessly. Thank you for taking the time to do this man.

jef:
Awesome! Thank you!

jef (approved):

PR #816, opened 2020-11-16
feat(notification): add product link to email and telegram

v1ct0rv:
Sometimes add to cart link does not work, so I'm sending also the product link.


jef (approved):
Awesome! Looks great. Thank you!

PR #832, opened 2020-11-18
feat(store): add amd rx 6000 series model

rcbevans:
### Description
AMD RX 6800/6800XT/6900XT to Series type.  Extend .env-example with new MAX_PRICE config.  Add AMD reference and Strix LC as model types.  Add Sapphire as new brand.

### Testing

Verified program runs without error.  Tested that filtering series to new AMD series runs, albeit there are no product pages to check.


frotofraggins:
was this added?

kckaiwei:
No, and the models have yet to be updated.

foxlet:
The URL for the regular RX 6800 (non-XT) is https://www.amd.com/en/direct-buy/5458373400/us by the way.

jef:
Looks great! Thanks for also sorting lexicographically!

jef (dismissed):

jef:
Sorry about the linting. I need to make an update to the max columns. I don't like it being this short.

tech-degenerate:
When will this be a feature we can get in a normal update? sorry sort of a noob...

jef (approved):

jeff4uraree:
6900xt link:  https://www.amd.com/en/direct-buy/5458372200/us

PR #860, opened 2020-11-20
fix(store): Move store check earlier to not try every store constantly

wburns:
### Description

Providing a store config does not prevent the store from being scheduled. This causes extra wake up processing and pollutes the logs when not needed. This allows for much easier debugging and seeing what stores are doing by removing logs for stores that weren't actually doing anything.

### Testing

Ran the application and verified the log message was printed only once per store and was not repeated, where as before it was repeated for every store every 10s.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,7 +7,7 @@ import {logger} from './logger';
 import puppeteer from 'puppeteer-extra';
 import resourceBlock from 'puppeteer-extra-plugin-block-resources';
 import stealthPlugin from 'puppeteer-extra-plugin-stealth';
-import {storeList} from './store/model';
+import {storeList, getStores} from './store/model';
 import {tryLookupAndLoop} from './store';
 
 puppeteer.use(stealthPlugin());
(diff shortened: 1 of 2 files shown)


jef:
@Doridian, this is pretty old now, but I think the logic is still sound (given we take updates into consideration).

I think last time I went to do this, the web UI portion broke. Would that still be the case here? I forget...

Doridian:
I don't really understand what "pollutes the logs" is, as only items are logged, not stores as such.
And with my proxy PR the link fetcher logic also got moved past that line, so those don't run anymore either for disabled stores.
This would still break the WebUI unless logic is added to start loops for stores if new stores are enabled. (and stop them if stores are disabled again), so we'd need better loop state management.

jef:
> I don't really understand what "pollutes the logs" is, as only items are logged, not stores as such.
> And with my proxy PR the link fetcher logic also got moved past that line, so those don't run anymore either for disabled stores.
> This would still break the WebUI unless logic is added to start loops for stores if new stores are enabled. (and stop them if stores are disabled again), so we'd need better loop state management.

This logic will be removed anyhow in 4.0.0 as it will grab from a live list, rather through stores. So it may not be worth the work either.

PR #862, opened 2020-11-20
feat(coolmod): add `maxPrice`

### LINKED ISSUE #835: Coolmod max_prixe not working
### Description

I have put a limit of 850 for the 3080. Is been working so far, but now is openning a coolmod window with the 3080, but the price is 1.006,95€.
https://www.coolmod.com/gigabyte-aorus-geforce-rtx-3080-xtreme-10gb-gddr6x-tarjeta-grafica-precio

Bug

egibar:
### Description

Fixes #835
The maxPrice was not added to the original .ts

### Testing

I have tested in a real enviroment, as there was one 3080 that was causing the bug in the coolmod store.

### DIFF
--- a/src/store/model/coolmod.ts
+++ b/src/store/model/coolmod.ts
@@ -6,6 +6,10 @@ export const Coolmod: Store = {
 			container: '.product-availability',
 			text: ['Envío']
 		},
+		maxPrice: {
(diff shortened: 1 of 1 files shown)


jef (approved):
Awesome! Thank you very much.

PR #872, opened 2020-11-21
chore: add more 3070 cards, reorder links

Barmaley13:
### Description
Added more 3070 links. Also reorganized them a bit so the series grouped together rather than scattered.

### Testing
Ran this version of the code on my machine and was able to purchase a card! Thank you so much for sharing this code with others!

### New dependencies
None


jef (approved):
Awesome! Thanks for helping maintain! Appreciate the help :smile:

PR #876, opened 2020-11-21
feat(store): add captcha check to saturn and mediamarkt

styler2go:
# Description

I have just added the captcha detection to the pages mediamarkt and saturn. Because they are from the same company the code is the same for both.


jef (approved):
Awesome! Thanks for the contribution!

PhriZ:
Is this still correct? 
I can't manage to repoduce any Captchas in Chrome no matter how often I refresh. 

I am probably missing something since I am a noob.

PR #878, opened 2020-11-22
feat(azerty): add ryzen 5000 series

RitoFaith:
### Description

Added ryzen 5000 series to azerty.ts

### Testing

Added all for of the ryzen 5000 series to azerty.ts and ran test for a couple of minutes, seems to work fine, 
sadly no stock though.


jef:
Thank you!

PR #889, opened 2020-11-22
fix(canadacomputers): add `inStock` selector, use english links

Sherloks:
### Description

In a new Chrome session, the canadacomputers website will use the system locale to set the store language. This will cause an issue if your System is in French. To prevent that, I added the language param in the URLs.

I also added "Online In Stock" selector, because on some product the store will let you pay online, but you can only pickup in store. In the current version it's useless to get an alert for that if you can't select your prefered store. I don't want to receive an alert for a store 2000km from my house.

### Testing

**To test the multilingual issue:** 
I guess you can try switching your Browser/System language (_lol_) or add `&language=fr` to the URLs

**To test the inStock/outOfStock issue:**
Find a product that has "Order Online and Pick Up In-Store".
At the moment there's 2x [Ryzen 5600x](https://www.canadacomputers.com/product_info.php?cPath=4_64_1969&item_id=183432&language=en) at Kanata, ON, but you can't Order Online and get it shipped.


src/store/model/canadacomputers.ts
```diff
@@ -2,9 +2,13 @@ import {Store} from './store';
...
 export const CanadaComputers: Store = {
 	labels: {
+		inStock: {
+			container: 'div[class="pi-prod-availability"]',
+			text: ['Online In Stock']
+		},
 		outOfStock: {
 			container: 'div[class="pi-prod-availability"]',
-			text: ['not available online']
+			text: ['Not Available Online']
```
jef, line 11: Casing won't matter here as we pull everything in and set it to lowercase.

jef (approved):
Looks great! Thanks for fixing. Good to keep in mind for other sites that use different locales.

PR #903, opened 2020-11-24
fix(newegg): typo in 5800x url

erikswain:
### Description

Original URL resolved properly due to Item# being correct, but my OCD noticed the URL was referencing the 5900x :)


jef (approved):
Ah, good catch! Thanks!

Issue #629, opened 2020-10-29
Pause lookup when in break time?


SesioN:
### Description

I've added for myself a feature, where you can define a break time, in which the bot does not lookup. 

Use-Case:
I run the bot on a server and only want notifications and thus cards getting lookup when I'm available. 
E.g. no need to spam the shops all night and it's annoying to start/stop the app manually.

Do we want this? We could in the future also support something like a schedule, especially in the WebApp.

### Possible solution
I haven't added a PR because my implementation was just a quick-dirty-thing. 

If we want this, I could refine the code a little. Maybe not only support military time. Also missing readme entry etc.

https://github.com/SesioN/nvidia-snatcher/commit/b2928a85b07a598fc8433fcf2a9f32b22eec53e6

jef:
I'd say it's a pretty good idea! Your implementation seems extensive; do you think creating a wrapper for the lookup with a `Date.now()` call around two variables (`BLOCK_START_TIME`, `BLOCK_END_TIME`) to see if we want to run the lookup?

PR #922, opened 2020-11-25
feat: option for max operating duration

ssweriduk:
### Description
Enables the ability to run this program for a specified period of time. This is especially useful when defining windows of operation.
For example, I want this program to only search for cards during the hours of 9AM -> 11PM as I don't want to be woken up in the middle of the night for card alerts. Setting a cron schedule to run daily at 9AM and a `MAX_RUNTIME` env var of `14400` achieves this. 
If I won't be actionable on the alerts this program sends out, why have it running. 
This is also useful for anybody looking to run this in a containerized environment in the cloud, and don't want to incur extra server costs during scheduled downtimes 

### Testing

Tested locally. Don't see any unit tests in this project, but happy to add some if needed!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -71,9 +71,11 @@ async function main() {
 	await startAPIServer();
 
 	if (config.maxRunTime) {
-		setTimeout(() => {
-			logger.info(`We've been searching for over ${config.maxRunTime} seconds as configured by MAX_RUNTIME. Time to call it quits for now. Gracefully stopping program.`);
-			stopAndExit();
+		setTimeout(async () => {
+			logger.info(
+				`We've been searching for over ${config.maxRunTime} seconds as configured by MAX_RUNTIME. Time to call it quits for now. Gracefully stopping program.`
+			);
+			await stopAndExit();
 		}, config.maxRunTime);
 	}
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -69,6 +69,13 @@ async function main() {
 	}
 
 	await startAPIServer();
+
+	if (config.maxRunTime) {
+		setTimeout(() => {
+			logger.info(`We've been searching for over ${config.maxRunTime} seconds as configured by MAX_RUNTIME. Time to call it quits for now. Gracefully stopping program.`);
+			stopAndExit();
+		}, config.maxRunTime);
+	}
 }
 
 async function stop() {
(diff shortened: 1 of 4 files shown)


src/store/lookup.ts
```diff
@@ -218,7 +218,7 @@ async function lookupCardInStock(store: Store, page: Page, link: Link) {
 }
 
 export async function tryLookupAndLoop(browser: Browser, store: Store) {
-	if (!browser.isConnected()) {
+	if (!browser || !browser.isConnected()) {
```
ssweriduk, line 221: `stopAndExit` sets `browser` to undefined. Adding this check here to gracefully exit this loop.

src/index.ts
```diff
@@ -69,6 +69,15 @@ async function main() {
...
 
 	await startAPIServer();
+
+	if (config.maxRunTime) {
+		setTimeout(async () => {
+			logger.info(
+				`We've been searching for over ${config.maxRunTime} seconds as configured by MAX_RUNTIME. Time to call it quits for now. Gracefully stopping program.`
+			);
+			await stopAndExit();
+		}, config.maxRunTime);
```
jef, line 79: This is in milliseconds, so you will have to convert.

jef (commented):
I could see how someone would want this feature, but then I could also see how this could be accomplished outside of the app. Perhaps even cause confusion (in terms of automation).

I'm just a little nervous in pulling this in, cause I don't know how much benefit it will add and perhaps cause confusion. But the change is so small, it shouldn't be a problem. So I'm not sure.

What do you think?

SesioN:
@ssweriduk 

I solved your issue in a different way, which should be better applicable.

https://github.com/jef/streetmerchant/issues/629

However I never got to refine this and make a PR. 

If you want, you could transform this into your own PR. I linked my dirty solution in the topic.

src/index.ts
ssweriduk: Doh! Thanks for catching.

ssweriduk:
@SesioN sorry for the delay! Holidays etc etc. Lemme take a look at that PR! Thanks for looking at this!

ssweriduk:
@jef Thanks for reviewing! And I agree, it could be confusing, I was struggling to come up with an env var name for this. I don't believe that AWS ECS has a way of sending `SIGTERM` to the containers on a schedule without setting up some sort of CloudWatch/Lambda integration, which I was trying to avoid. Open to other solutions!

src/store/lookup.ts
jef: Oh! I think this could help #1174. Not sure though, haven't looked into in depth.

jef (changes requested):
After we fix that up, I think we can pull this in.

Thanks!

PR #926, opened 2020-11-25
chore(amd): add rx6800

foxlet:
### Description

Adds the RX 6800 (non-XT) card from the AMD Direct store; only the 6800XT is available currently.

### Testing

1. Ran tracking with `STORES=amd` and `SHOW_ONLY_SERIES=rx6800`.
2. Verified `OUT OF STOCK` response.


jef (approved):
Great! Thank you.

PR #933, opened 2020-11-25
feat(proxy): socks5 support

joslynesser:
### Description

Currently, only `http` proxies are supported. By adding the ability to specify a protocol, one can use a secured `socks5` proxy instead. One good use case for this is tunneling through SSH to a remote server to avoid making connections from your home IP. Default fallback is `http`, leaving backwards compatibility with the current static default of `http`.

### Testing

Tested with no protocol specified, which results in current behavior of using `http`. 

After setting `PROXY_PROTOCOL=socks5`, puppeteer now connects through `SOCKS5` to my `PROXY_ADDRESS:PROXY_PORT`.


jef (approved):
Awesome! Great addition. Thanks for adding.

PR #938, opened 2020-11-26
feat(store): add 6800 series Pccomponentes & Coolmod

Crstian19:


jef:
I don't think all of these cards are `amd reference` cards. They would be like `nitro plus` or something like that.

jef (changes requested):
I don't think all of these cards are `amd reference` cards. They would be like `nitro plus` or something like that.

PR #952, opened 2020-11-26
chore(store): add 6800 & 6800xt cards to overclockers

eddwatts:
improvement(store): add 6800 & 6800xt cards to overclockers


jef (approved):
Thank you!

PR #954, opened 2020-11-26
fix: related products being picked up on scan

reesvarney:
### Description
There isn't an issue for this as its so small but related products were being picked up on scan. I've just changed the css selector to filter out these.

Just changed `ul.productColumns` to `ul.productColumns:not(.related)`.

### Testing

Tests weren't really necessary for this, related products were no longer picked up by the scraper. I'm not super confident in writing them but I can give it a try if its that necessary.


jef:
Great! Thanks!

jef (approved):

PR #974, opened 2020-11-27
chore(store): add amd-it store and RX 6800 XT

MartelliEnrico:
Added the italian amd store and the amd radeon rx 6800 xt to the listing.


jef (approved):
Great! Thanks!

PR #1002, opened 2020-11-28
feat(store): update the italian store links

lucarin91:
### Description
This PR adds new links for the Italian stores (i.e., amazon-it, unieuro, euronics, and comet). In particular, it adds almost all the GPU and CPU on amazon-it, and the Microsoft console on the other stores.

### Testing
I check the validity of each link.


jef (approved):
Great! Thank you!

PR #1008, opened 2020-11-28
fix: Fixing bestbuy-ca.ts in-stock checker 

LazG7:
### Description
Was broken as it looks like the bestbuy.ca website was updated since this store was added. 

### Testing
Tested lines 4 through 22 by running test:brand, test:model and got a successful return on it.

### New dependencies
None

### DIFF
--- a/src/store/model/bestbuy-ca.ts
+++ b/src/store/model/bestbuy-ca.ts
@@ -3,8 +3,8 @@ import {Store} from './store';
 export const BestBuyCa: Store = {
 	labels: {
 		inStock: {
-			container: '.x-product-detail-page .addToCartButton:enabled',
-			text: ['Add to Cart']
+			container: '.pricingContainer_25k3c',
+			text: ['href="http://schema.org/InStock"']
 		},
 		maxPrice: {
 			container:
@@ -18,7 +18,7 @@ export const BestBuyCa: Store = {
 			model: 'test:model',
 			series: 'test:series',
 			url:
-				'https://www.bestbuy.ca/en-ca/product/msi-nvidia-geforce-rtx-2060-super-gaming-x-8gb-gddr6-video-card/14419420?intl=nosplash'
+				'https://www.bestbuy.ca/en-ca/product/hp-deskjet-3755-wireless-all-in-one-inkjet-printer-seagrass-only-at-best-buy/10747797?intl=nosplash'
 		},
 		{
 			brand: 'zotac',
--- a/src/store/model/amazon-jp-en.ts
+++ b/src/store/model/amazon-jp-en.ts
@@ -1,0 +1,43 @@
+import {Store} from './store';
+
+export const AmazonJpEN: Store = {
+    backoffStatusCodes: [403, 429, 503],
+    labels: {
+        captcha: {
+            container: 'body',
+            text: ['enter the characters you see below']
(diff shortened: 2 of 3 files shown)


LazG7:
In my effort (and eventual screw up) to add a store I managed to make the mistake of adding it to this pull request. It does fix bestbuy-ca but I've also managed to include adding amazon-jp-en for english speaking people in japan (as well as the edits to the index.ts) trying to acquire the ps5c or ps5de.

I've tested this and it 100% works (is looking through the english version of the page for add to cart) using test:series.

Sorry sir! Let me know if there's something I can do to fix this or if these errors are nonsense. Still have a lot to learn with github. Trying to contribute and made a mess...

src/store/model/bestbuy-ca.ts
```diff
@@ -3,8 +3,8 @@ import {Store} from './store';
 export const BestBuyCa: Store = {
 	labels: {
 		inStock: {
-			container: '.x-product-detail-page .addToCartButton:enabled',
-			text: ['Add to Cart']
+			container: '.pricingContainer_25k3c',
```
e-rap, line 6: What was broken with the current css selector? I just did a check with chrome developer tools and it seems to be a valid check for in stock items.

src/store/model/bestbuy-ca.ts
```diff
@@ -3,8 +3,8 @@ import {Store} from './store';
...
+			text: ['href="http://schema.org/InStock"']
```
e-rap, line 7: Looking into href="http://schema.org/InStock" it seems there are other valid states when look at different products on bestbuy.ca [https://schema.org/ItemAvailability](https://schema.org/ItemAvailability) such as "http://schema.org/OnlineOnly"

src/store/model/bestbuy-ca.ts
LazG7: Sitting in the discord support chat a user was getting false positives because it was displaying "add to cart" in a gray box. It appears they rewrote the websites code (I think) since the original bestbuy-ca.ts was made and now every item (unless the schema.org is different) if it's out of stock it doesn't say, it just turns the button gray as opposed to saying out of stock or add to cart.

src/store/model/bestbuy-ca.ts
LazG7: I'm not certain but is it possible to have more than one `text: [ ]` to search for?  Is it possible to just comma separate them and add more?

src/store/model/bestbuy-ca.ts
rdfurman: I noticed some strange-ness about the BestBuy checker as well. But for me it was that the add to cart button initially loads in enabled, some lazy load JavaScript comes in and check stock, then is subsequently disables the button. Problem was that the scraper checked the button BEFORE it was disabled.

rdfurman: Almost seems like we need some kind of algo to delay the check of "out of stock" by a env amount. If we delay the button enable status check by say 3 seconds this would be a non-issue.

xCoBaLTz: This selector wont work. The `.pricingContainer_####` class is appended with a hash, meaning it's going to be different for every item.

Issue #882, opened 2020-11-22
Twilio Error with no values entered in .env


AlphaOmega2020:
### Description

**Error**
```
[~] # docker run --cap-add=SYS_ADMIN -it --rm --env-file /share/Container/container-station-data/appdata/streetmerchant/config/.env ghcr.io/jef/streetmerchant:latest

> streetmerchant@ start:production /app
> node build/index.js

ＳＴＲＥＥＴＭＥＲＣＨＡＮＴ
3.0.0

/app/node_modules/twilio/lib/rest/Twilio.js:140
    throw new Error('accountSid must start with AC');
    ^

Error: accountSid must start with AC
    at new Twilio (/app/node_modules/twilio/lib/rest/Twilio.js:140:11)
    at Object.<anonymous> (/app/build/notification/twilio.js:10:14)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/app/build/notification/notification.js:17:18)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! streetmerchant@ start:production: `node build/index.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the streetmerchant@ start:production script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/appuser/.npm/_logs/2020-11-22T12_10_58_942Z-debug.log

```

**.env**
```
# ** All configuration variables are optional **
# Read https://github.com/jef/streetmerchant#customization for help on customizing this file
#############################################################################################

ASCII_BANNER=""
ASCII_COLOR=""
AUTO_ADD_TO_CART="true"
BROWSER_TRUSTED=""
COUNTRY="UK"
DESKTOP_NOTIFICATIONS=""
DISCORD_NOTIFY_GROUP=""
DISCORD_WEB_HOOK=""
EMAIL_PASSWORD=""
EMAIL_TO=""
EMAIL_USERNAME=""
[...]

Issue #807, opened 2020-11-15
Microcenter locations


Freddynyc:
### Description

Microcenter will not give me alerts for CPUs in stock at specific locations.  It will only give alerts if I leave the field blank to default it.

```
[9:48:30 AM] info :: ✖ [microcenter] [amd (ryzen5600)] 5600x :: OUT OF STOCK
[9:48:30 AM] info :: ✖ [newegg] [amd (ryzen5900)] 5900x :: OUT OF STOCK
[9:48:31 AM] info :: ✖ [amd] [amd (ryzen5600)] 5600x :: OUT OF STOCK
[9:48:31 AM] info :: ✖ [newegg] [amd (ryzen5800)] 5800x :: OUT OF STOCK
[9:48:33 AM] info :: ✖ [bestbuy] [asus (3080)] strix :: OUT OF STOCK
[9:48:33 AM] info :: ✖ [microcenter] [amd (ryzen5800)] 5800x :: OUT OF STOCK
[9:48:36 AM] info :: ✖ [newegg] [amd (ryzen5600)] 5600x :: OUT OF STOCK
[9:48:37 AM] info :: ✖ [microcenter] [amd (ryzen5900)] 5900x :: OUT OF STOCK
[9:48:41 AM] info :: ✖ [amd] [amd (ryzen5900)] 5900x :: OUT OF STOCK
[9:48:41 AM] info :: ✖ [bestbuy] [amd (ryzen5900)] 5900x :: OUT OF STOCK
[9:48:41 AM] info :: ✖ [bandh] [amd (ryzen5900)] 5900x :: OUT OF STOCK
[9:48:41 AM] info :: ✖ [microcenter] [amd (ryzen5600)] 5600x :: OUT OF STOCK
[9:48:43 AM] info :: ✖ [amd] [amd (ryzen5800)] 5800x :: OUT OF STOCK
[9:48:44 AM] warn :: ✖ [bandh] [amd (ryzen5800)] 5800x :: STATUS CODE ERROR 503
[9:48:45 AM] info :: ✖ [amazon] [asus (3080)] strix oc :: OUT OF STOCK
[9:48:47 AM] info :: ✖ [microcenter] [amd (ryzen5800)] 5800x :: OUT OF STOCK
[9:48:48 AM] info :: ✖ [amd] [amd (ryzen5600)] 5600x :: OUT OF STOCK
[...]

PR #1011, opened 2020-11-29
docs: updated README.md for .env and docker issues

Spartanaco:
### Description
Addresses issues when getting streetmerchant running in Docker encountered by myself and others when using the `--env-file` option with `docker run`.

Relates to #882 and #807 

### Testing
I ran a few docker run commands illustrating the types of errors seen when you either leave blank options unset in the `.env` file or wrapped certain setting values in double quotes (`"`).

**E.g.**

* Wrapping the `STORES` setting value in double quotes like `STORES="adorama,amazon,amd,asus,bandh,bestbuy,gamestop,microcenter,newegg,officedepot,target,walmart,zotac"` results in the following `warn` on startup stating `No store named zotac", skipping.` with the command `docker run --cap-add=SYS_ADMIN --name=streetmerchant -it --rm --env-file ./.env ghcr.io/jef/streetmerchant:latest`:
image

* Also wrapping the `MICROCENTER_LOCATION` setting value in double quotes, like `MICROCENTER_LOCATION="denver"` results in the following error when starting with `docker run --cap-add=SYS_ADMIN --name=streetmerchant -it --rm --env-file ./.env ghcr.io/jef/streetmerchant:latest`:
image

* Lastly there's odd startup errors regarding certain notification components (like Twilio) when the optional settings are left blank in the `.env` file, as copied from the `.env-example` file like so:
```
TWILIO_ACCOUNT_SID=""
TWILIO_AUTH_TOKEN=""
TWILIO_FROM_NUMBER=""
TWILIO_TO_NUMBER=""
```
Which results in the following error on startup when using Docker with `docker run --cap-add=SYS_ADMIN --name=streetmerchant -it --rm --env-file ./.env ghcr.io/jef/streetmerchant:latest`:
image

The solution for me was to remove all wrapping double quotes from any setting values used in the `.env` file as well as removing any settings that I was not intending to use from the `.env` file. So I feel like some clarifications in the README.md are warranted to help avoid confusion for newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,9 @@ filesystem changes are detected in the `src/` folder or `.env` file.
 
 To customize `streetmerchant`, make a copy of `.env-example` as `.env` and make any changes to your liking. _All
 environment variables are **optional**._
+> :page_facing_up: Don't copy empty or otherwise unused options to your `.env` as it can cause issues when running in Docker and passing in the `.env` file with the `--env-file` option.
+
+> :page_facing_up: Also don't wrap your settings values in double quotes (`"`) in the `.env` file when using Docker, as this can also cause issues.
 
 <details>
 <summary>Expand to see all available options</summary>


jef:
I think this has to do more with parsing and it should be fixed. I'm going to go through several stages of config changes (tomorrow?) and this will hopefully address is.

I really appreciate you updating the doc, but it bandages up with what's wrong with the code 😢

Spartanaco:
Thanks for the response @jef , I appreciate it, and I agree it's better to fix the underlying parsing issue than to accommodate odd behavior with the README.md. I will close this now.

PR #1016, opened 2020-11-29
feat(store): add ePrice store

lucarin91:
### Description
I add the [ePrice](https://www.eprice.it/) with the link for the series 3070 and the PS5.
This store now sells PS5 at over 1000€ 😱, thus I also added the filter variable for the PS5 price.
Finally, I fix an error in the `.env-example` file in which the `MAX_PRICE_SERIES_CORSAIR_SF` variable was written as `MAX_PRICE_CORSAIR_SF`.
This time I also update the `README`. 

P.S. sorry for the PR spam 😅 .

### Testing
As with the other PRs, I check the code with in-stock items in the store.


jef (approved):
Great! Thanks for adding those max prices!

PR #1026, opened 2020-11-30
feat: add proxy rotation (settable per store)

Doridian:
### Description

Adds ability to add multiple proxies per store, rotated once per fetch.
Proxies are read from a file `STORENAME.proxies` in the format like `socks5://username:password@ip` one per line.
Adds some require architectural changes for that to ensure adblocker doesn't explode (puppeteer only supports one request listener per instance, so we have to write our own chain there)


jef:
@Doridian with the great contribution! Thanks for help. I'll take a look at this tonight and get back.

MujiSayed:
quick question about this. I know format is  socks5://username:password@ip, but what if there is no username/password for said socks5 proxies? Also since default port is 1080, is that what it's guessing as it's port or will a format of socks5://ip:port work as well?

Thanks for the contribution!

Doridian:
Correct. That format should work. Default port is 1080 indeed.

src/store/lookup.ts
```diff
@@ -9,18 +9,109 @@ import {
...
+
+	return store.proxyList[store.currentProxyIndex];
+}
+
+async function handleLowBandwidth(request: Request) {
+	if (!config.browser.lowBandwidth) {
+		return false;
+	}
+
+	const typ = request.resourceType();
```
jef, line 51: Is this because `type` is a reserved word? Haha :stuck_out_tongue_closed_eyes:

jef (approved):
Amazing. Thank you so much @Doridian. This really needed some clean up. I appreciate the time spent on the refactoring as well!

Doridian:
Indeed it is. 

On Tue, Dec 1, 2020, at 16:11, Jef LeCompte wrote:
> 

> *@jef* approved this pull request.

> Amazing. Thank you so much @Doridian <https://github.com/Doridian>. This really needed some clean up. I appreciate the time spent on the refactoring as well!

> 
> In src/store/lookup.ts <https://github.com/jef/streetmerchant/pull/1026#discussion_r533806329>:

> > +		store.currentProxyIndex = 0;
+	}
+
+	logger.info(
+		`ℹ [${store.name}] Next proxy index: ${store.currentProxyIndex} / Count: ${store.proxyList.length}`
+	);
+
+	return store.proxyList[store.currentProxyIndex];
+}
+
+async function handleLowBandwidth(request: Request) {
+	if (!config.browser.lowBandwidth) {
+		return false;
+	}
+
+	const typ = request.resourceType();

JasonMore:
@Doridian which socks5 proxies do you use? Looking for suggestions

gigi2006:
@Doridian @jef 
would also be very interested which proxys you recommend...

PR #1045, opened 2020-11-30
feat(store): add Euronics DE

fnk93:
### Description
Add Euronics Germany store support. Checks for XBox Series S


jef (approved):
You're on a roll! Thank you!

PR #1046, opened 2020-11-30
feat(store): add Gamestop DE

fnk93:
### Description
Add support for Gamestop DE. Allow checking of PS5 disc and digital version availability.


jef (approved):
Yahoo! Thank you!

PR #1056, opened 2020-12-01
feat(store): add spielegrotte

fnk93:
### Description
Add support for Spielegrotte. Allow checking of PS5 disc and digital version availability.

### DIFF
--- a/src/store/model/spielegrotte.ts
+++ b/src/store/model/spielegrotte.ts
@@ -1,0 +1,44 @@
+import {Store} from './store';
+
+export const Spielegrotte: Store = {
+	labels: {
+		inStock: [
+			{
+				container:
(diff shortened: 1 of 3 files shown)


src/store/model/spielegrotte.ts
```diff
@@ -0,0 +1,44 @@
+import {Store} from './store';
+
+export const Spielegrotte: Store = {
+	labels: {
+		inStock: [
+			{
+				container:
+					'html > body > table > tbody > tr > td > div > table > tbody > tr > td > center > table > tbody > tr > td > a.klein > img',
```
jef, line 8: I'm thinking this might be a better selector

```suggestion
					'img[src="neu_indenkorb_2.gif"]',
```

jef: Actually, scratch that. Let's keep it like the others.

PR #1057, opened 2020-12-01
feat(store): add Medimax

fnk93:
### Description
Add support for Medimax. Allow checking of PS5 disc and digital version availability.


jef (approved):
You're a roll! Thank you!

PR #1065, opened 2020-12-01
chore(cyberport): remove xbox/ps5

fnk93:
### Description
During update process wrong urls have been pasted into cyberport. Removed them as I currently know of no active urls.


jef (approved):
Thank you. I really missed this one.

PR #1074, opened 2020-12-01
feat(bestbuy): add 3060 series

petertdinh:


petertdinh:
bestbuy is the only page so far that seems to be showing stock, i'll add more links if ppl could find other pages.

jef (approved):
Sweet! Thank you!

PR #1085, opened 2020-12-02
chore(bestbuy): Add 3060ti Models

MrMcChewy:


jef (changes requested):
Some of these cards are already added. Do you mind removing the duplicates and we can pull this in!

PR #1096, opened 2020-12-02
feat(store): add Otto

fnk93:
### Description
Add support for Otto. Allow checking of PS5 disc and digital version availability, as well as XBox Series X and S.
Two PS5 disc version urls are included as they take the sale url offline once the item is sold out, so I couldn't test out which is the current url.


jef (approved):
Awesome. Thanks!

PR #1100, opened 2020-12-02
chore(store): update eu stores

fnk93:
### Description
Fix euroFormat for several EU stores.
Add Xbox Series X, PS5 disc/digital to Euronics (DE).
Add PS5 disc bundle to Alternate.
Change test item on several EU Amazon sites to a product available.

### DIFF
--- a/src/store/model/euronics-de.ts
+++ b/src/store/model/euronics-de.ts
@@ -8,20 +8,52 @@ export const EuronicsDE: Store = {
 		},
 		maxPrice: {
 			container: '.price--content',
-			euroFormat: true
+			euroFormat: false
 		},
 		outOfStock: {
-			container: '.product--buybox .alert--content',
-			text: ['Artikel steht derzeit nicht zur Verfügung']
+			container:
+				'.product--buybox .alert.is--error.is--rounded .alert--content',
+			text: [
+				'Artikel steht derzeit nicht zur Verfügung',
+				'Morgen im Laufe des Morgens und nur online unter'
+			]
 		}
 	},
 	links: [
+		{
+			brand: 'test:brand',
+			model: 'test:model',
+			series: 'test:series',
+			url:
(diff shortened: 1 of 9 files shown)


src/store/model/euronics-de.ts
```diff
@@ -8,20 +8,52 @@ export const EuronicsDE: Store = {
...
+			]
 		}
 	},
 	links: [
+		{
+			brand: 'test:brand',
+			model: 'test:model',
+			series: 'test:series',
+			url:
+				'https://www.euronics.de/telefon-und-navigation/festnetz/schnurlose-telefone/kx-tg6721gb-schnurlostelefon-mit-anrufbeantworter-schwarz-4051168442801'
```
jef, line 28: It seems that this content is in fact in `euroFormat`

`31,16 €`

jef (changes requested):
Could you explain a little more on the euroformat changes? Thanks!

src/store/model/euronics-de.ts
fnk93: Just for my understanding, as I think I may have misunderstood the use of euroFormat.
Is it to toggle between "." for true and "," for false as separator between the thousands?
If yes, then give me some minutes, I'll fix the mess I did here.

joerg-knitter:
Why does the Alternate link refer to the MEGA bundle of the PS5 console, not the "standard" edition without bundled games?

fnk93:
@joerg-knitter I only added the mega bundle, the standard disc url has been in already before and is still in there: https://www.alternate.de/product/1651220

src/store/model/euronics-de.ts
jef: Yep! That's exactly it. Thanks for fixing and also updating old ones 😄

jef (approved):
~I think the containers in `euronics-de.ts` are wrong? But that's not your doing. I'll have to go in and test.~ It's fine actually. Sorry about that!

Thanks!

fnk93:
Already have been panicking and looking for the html source to find my mistake.

PR #1102, opened 2020-12-02
feat(store): add amazon-de-warehouse

fnk93:
### Description
Add support for amazon (de) warehouse. Allow checking of PS5 disc and digital version availability, as well as XBox Series X and S.

### DIFF
--- a/src/store/model/amazon-de-warehouse.ts
+++ b/src/store/model/amazon-de-warehouse.ts
@@ -1,0 +1,63 @@
+import {Store} from './store';
+
+export const AmazonDeWarehouse: Store = {
+	backoffStatusCodes: [403, 429, 503],
+	labels: {
+		captcha: {
+			container: 'body',
+			text: [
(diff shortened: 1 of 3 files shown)


src/store/model/amazon-de-warehouse.ts
```diff
@@ -0,0 +1,63 @@
...
+				]
+			}
+		]
+	},
+	links: [
+		{
+			brand: 'test:brand',
+			model: 'test:model',
+			series: 'test:series',
+			url: 'https://www.amazon.de/gp/offer-listing/B087VM5XC6'
```
jef, line 35: Do you mind changing this to something that may not go out of stock for awhile?

jef (changes requested):

src/store/model/amazon-de-warehouse.ts
fnk93: Of course, give me a few minutes

jef (approved):
Thanks!

PR #1105, opened 2020-12-02
feat(store): add expert

fnk93:
### Description
Add support for expert. Allow checking of PS5 disc and digital version availability, as well as XBox Series S.


jef (approved):
Great! Thank you very much!

PR #1107, opened 2020-12-02
feat(store): add el corte ingles

fnk93:
### Description
Add El Corte Ingles with PS5 and Xbox Series X/S support.

### DIFF
--- a/src/store/model/elcorteingles.ts
+++ b/src/store/model/elcorteingles.ts
@@ -1,0 +1,69 @@
+import {Store} from './store';
+
+export const Elcorteingles: Store = {
+	backoffStatusCodes: [403, 429, 503],
+	labels: {
+		// Captcha: {
+		// 	container: 'body',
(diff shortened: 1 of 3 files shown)


src/store/model/elcorteingles.ts
```diff
@@ -0,0 +1,69 @@
...
+export const Elcorteingles: Store = {
+	backoffStatusCodes: [403, 429, 503],
+	labels: {
+		// Captcha: {
+		// 	container: 'body',
+		// 	text: [
+		// 		'geben sie die unten angezeigten zeichen ein',
+		// 		'geben sie die zeichen unten ein'
+		// 	]
+		// },
```
jef, line 12: Is this not good? Or does it not come up enough?

jef (approved):
Let's merge this in, but maybe we can figure out captcha later or delete the comment.

Thanks!

PR #1108, opened 2020-12-02
fix(store): update euroFormat for expert and otto

fnk93:
### Description
Messed up with the euroFormat value for Expert and Otto which have already been pulled. This fixes the issue.


jef (approved):
Thank you for the update 😃

PR #1110, opened 2020-12-02
fix(store): add backoffStatusCodes and new captcha container

fnk93:
### Description
Add backoffStatusCodes to Expert, Mediamarkt and Saturn.
Change captcha container for Mediamarkt and Saturn as it wasn't recognized before. Maybe this is solving this.

### DIFF
--- a/src/store/model/expert.ts
+++ b/src/store/model/expert.ts
@@ -1,6 +1,7 @@
 import {Store} from './store';
 
 export const Expert: Store = {
+	backoffStatusCodes: [403, 429, 503],
 	labels: {
 		inStock: [
 			{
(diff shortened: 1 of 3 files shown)


jef (approved):
Thank you!

joerg-knitter:
I think, expert.de uses an IP blocker in case that there are too many requests within a short time. With a chrome extension (before I learned to know streetmerchant), I had a check at around every 10s, and I got a "forbidden" for around 7 days on every device in the house while I was able to access the site with a Tor Browser.
So a delay might be necessary in addition to some captchas that I haven´t seen yet due to the blocking ;)

PR #1126, opened 2020-12-03
fix(store): removes unnecessary newegg links from bestbuy

### LINKED ISSUE #1119: Bestbuy.ts contains cards and urls that point to Newegg.com
### Description

Using Best Buy searches for cards on Newegg
This is a Big Navi specific problem and should be an easy fix

### Possible solution

Remove all entries that aren't for best buy. The only AIB card that Best Buy sells is the XFX Merc.

jaw20:
### Description
This fixes #1119 and removes the extraneous links specific to the newegg.ts file that are in the bestbuy.ts file. This is required so that people don't accidentally ban themselves from Newegg when they were trying to only search Bestbuy. This is my second attempt at this pull request so as to conform to the commit styling guidelines.


jef:
Oh no... I did this! I did a huge goof. Thanks for fixing. Do you know if these stores are in newegg?

jef (approved):

PR #1145, opened 2020-12-04
fix(bestbuy-ca): model always indicating in stock

### LINKED ISSUE #1093: All cards notified as InStock for BestBuy.ca after last update
Hello,

all the cards (3070,3080) are notified as in stock even though there are not on bestbuy-ca.

My config didn't change and it was working yesterday I just did a pull request this morning and it stopped working :/

Thanks for the help!

SamP301:
### Description

Fixes #1093

It seams that the page changed to always include the "Add to cart button". However, when there is no card in stock, that button is disabled. I used the class applied on the button when there is no items in stock to fix the issue.

### Testing

I tried with the test URL and with some cards that where not in stock. The same class is added to the button when the product is not in stock.

Hope it helps !

### DIFF
--- a/src/store/model/bestbuy-ca.ts
+++ b/src/store/model/bestbuy-ca.ts
@@ -1,170 +1,170 @@
-import { Store } from "./store";
+import {Store} from './store';
 
 export const BestBuyCa: Store = {
 	labels: {
 		inStock: {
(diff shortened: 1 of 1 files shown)


jef (approved):
Awesome! Yeah, I think this will work! Thank you.

PR #1157, opened 2020-12-04
docs: add cd step

srolandmarshall:
added step for navigating to the folder, for newbies


jef (approved):
Definitely seems to be a common thing to miss this step.

Thanks!

Issue #1194, opened 2020-12-05
Screenshot


DerHOgi:
For a few tests:
Is it possible to take a screenshot everytime and not only on success?
What do I have to change in the code?

jef:
This should do it for you.

```diff
diff --git a/src/store/lookup.ts b/src/store/lookup.ts
index 275557a..8082177 100644
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -284,13 +284,13 @@ async function lookupCard(
 				inStock[link.url] = false;
 			}, 1000 * config.page.inStockWaitTime);
 		}
+	}
 
-		if (config.page.screenshot) {
-			logger.debug('ℹ saving screenshot');
+	if (config.page.screenshot) {
+		logger.debug('ℹ saving screenshot');
 
-			link.screenshot = `success-${Date.now()}.png`;
-			await page.screenshot({path: link.screenshot});
-		}
+		link.screenshot = `success-${Date.now()}.png`;
+		await page.screenshot({path: link.screenshot});
 	}
 
 	return statusCode;
```

DerHOgi:
Thanks @jef for you reply!
Now I can see what's going on and why I allways get false negative results. Item is available but bot result is OUT OF STOCK!
![success-1607193640887](https://user-images.githubusercontent.com/67831862/101260868-22b3ca80-3733-11eb-9f21-c0b1192a99b6.png)
What can we do against this. All amazon-*.ts sites!

jef:
I actually just fixed this bug too! Not sure how it ever happened, but #1209

gigi2006:
> Thanks @jef for you reply!
> Now I can see what's going on and why I allways get false negative results. Item is available but bot result is OUT OF STOCK!
> What can we do against this. There is a new captcha-check that concerns all Europe amazon-*.ts sites!
> ![success-1607193640887](https://user-images.githubusercontent.com/67831862/101260868-22b3ca80-3733-11eb-9f21-c0b1192a99b6.png)

hey @jef 

how do you think we can solve the problem with the captcha? i don't think there will be a perfect solution, i already set and tried PAGE_BACKOFF_MIN and PAGE_BACKOFF_MAX as well as PAGE_SLEEP_MIN and PAGE_SLEEP_MAX. But it did not help much.

now a consideration, couldn't we just let the browser open when the captcha warning appears and then the real human could solve the captcha from time to time? so i would do it.

@DerHOgi 
how could you solve the problem?
[...]

PR #1212, opened 2020-12-06
fix(store): italian store checks

lucarin91:
### Description
Fix some issues on some Italian store.
In particular:
- ePrice store sometimes products where show wrongly in stock
- Amazon-it do not correctly see the caption page
- Uniero return a blank page after too many requests

### Test
I used the trick on issue #1194 to always save a screenshot. Maybe it can be useful to have a configuration option to do that.

### DIFF
--- a/src/store/model/amazon-it.ts
+++ b/src/store/model/amazon-it.ts
@@ -5,7 +5,9 @@ export const AmazonIt: Store = {
 	labels: {
 		captcha: {
 			container: 'body',
-			text: ['enter the characters you see below']
+			text: [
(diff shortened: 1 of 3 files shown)


jef:
> I used the trick on issue #1194 to always save a screenshot. Maybe it can be useful to have a configuration option to do that.

Good point. Maybe I can add this in. Except it would start taking up a lot of disk space quickly, so perhaps rolling images.

jef (approved):
Thanks for the updates and fixes!

razor992fg:
@lucarin91 hello luca, can you add mediaworld.it please??? i'm italian :)

PR #1229, opened 2020-12-06
chore(amazon): add 3060ti cards

dsmaugy:
### Description

added support for 3060ti cards on amazon on amazon.ts 

### Testing

Only modified src/store/model/amazon.ts

### New dependencies

NA


jef (approved):
Looks great! Thanks!

Issue #1237, opened 2020-12-07
Failing to parse prices for amazon-it store


Zimm1:
## Current Behavior

`amazon-it` prices don't get parsed correctly because it's using the wrong decimal separator.

## Environment

**.env file:**

```dotenv
STORES="amazon-it"
```

Issue #1238, opened 2020-12-07
Wrong url for Asus 3070 Dual in amazon-it store


Zimm1:
## Current Behavior

A current url for an Asus 3070 Dual points to the Fire TV stick.
`https://www.amazon.it/gp/aws/cart/add.html?ASIN.1=B07PW9VBK5&Quantity.1=1`

## Environment

**.env file:**

```dotenv
STORES="amazon-it"
```

PR #1239, opened 2020-12-07
fix(amazon-it): maxPrice selector and links

Zimm1:
### Description

Fixes issues #1237 #1238 

The `amazon-it` store is now updated to fix some bugs

- Fixed price parsing using the correct decimal separator
- Fixed an Asus 3070 Dual url, which pointer to an Amazon Fire Stick instead


jef (approved):
Great! Thank you very much.

PR #1278, opened 2020-12-09
fix(bestbuy-ca): update selectors to help false positives

dylanfontaine:
### Description

* Noticed the bestbuy ca selectors were using generated class names, switched it to be more generic in case those class names change


src/store/model/bestbuy-ca.ts
```diff
@@ -2,17 +2,13 @@ import {Store} from './store';
 
 export const BestBuyCa: Store = {
 	labels: {
-		inStock: {
-			container: '.addToCartContainer_2uzan',
-			text: ['add to cart']
-		},
```
jef, line 8: I think we'll want to keep some form of `inStock`. This is good for false positive detection.

jef (approved):
Let's merge this in and go from there.

Thanks!

PR #1299, opened 2020-12-09
chore(amazon-ca): add rtx 3070 cards

SizzlinSeal:
### Description

added some 3070 cards on amazon-ca

### Testing

I built it and it worked. Then I used it and it worked.

I also tested all the links manually


jef (approved):
Thank you very much! Great work.

Issue #1035, opened 2020-11-30
Desktop notifications need appID set in Windows 10


Hedova928:
### Description

Desktop notifications through node-notifier are not working in Windows 10. This seems to be due to the requirement to set an `appID`: https://www.npmjs.com/package/node-notifier#usage-windowstoaster

```
C:\Users\Hedova\Desktop\Programs\streetmerchant>npm run test:notification

> test:notification
> npm run build && node build/__test__/notification-test.js


> build
> tsc

ＳＴＲＥＥＴＭＥＲＣＨＡＮＴ
3.1.0

INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
[4:31:26 AM] info :: ✔ sound player found: cmdmp3
[4:31:27 AM] info :: ✔ desktop notification sent
Notifications are disabled
Reason: DisabledForUser Please make sure that the app id is set correctly.
Command Line: C:\Users\Hedova\Desktop\Programs\streetmerchant\node_modules\node-notifier\vendor\snoreToast\snoretoast-x64.exe -appID streetmerchant -pipeName \\.\pipe\notifierPipe-efc1ad6d-7071-4038-966f-ffe741f32252 -m https://www.example.com/cartUrl -t "[4:31:29 AM] info :: ✔ played sound
```

jef:
Interesting... Thanks for the heads up

> With the Fall Creators Update, Notifications on Windows 10 will only work as expected if a valid `appID `is specified. Your `appID `must be exactly the same value that was registered during the installation of your app.

In this case, I'm unsure what the `appID` would be. I don't know if we register it or not.

PR #1301, opened 2020-12-09
feat(notification): add sound player executable option

warmar:
### Description

Allows the default sound player to be overridden. I was having trouble on Ubuntu with the default player which sound-player chose.

### Testing

Ran locally

### New dependencies

None

### DIFF
--- a/docs/reference/notification.md
+++ b/docs/reference/notification.md
@@ -8,6 +8,7 @@ You can test your notification configuration by running `npm run test:notificati
 |---|---|
 | `DESKTOP_NOTIFICATIONS` | Display desktop notifications using [node-notifier](https://www.npmjs.com/package/node-notifier). |
 | `PLAY_SOUND` | Play this sound notification if a product is found. Relative path accepted, valid formats: wav, mp3, flac, E.g.: `path/to/notification.wav`, [free sounds available](https://notificationsounds.com/) |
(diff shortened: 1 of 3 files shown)


jef:
Cool! Looks good to me. Quick question though, what sound player do you have? It seems like the supported libraries are sufficient, but maybe that's not the case.

warmar:
> Cool! Looks good to me. Quick question though, what sound player do you have? It seems like the supported libraries are sufficient, but maybe that's not the case.

By default, play-sound detects aplay, which is indeed installed on my system. When it plays a sound, however, all I hear is static, buzzing, and popping. Even when I run aplay from the command line, this happens. I have not yet been able to identify the problem.

For now, I just switched to using vlc and that has been working fine.

jef (approved):

jef:
> Looks like installing sox (play) also resolved the issue. I can still see this option being useful on systems with several of the default supported players installed.

I also use `sox` on my system, so I never ran into that one before! Thanks for the clarification.

I agree, the default list is somewhat limiting _and_ it also causes the behavior on Windows to post `INFO: Could not find files for the given pattern(s).`

See related: https://github.com/jef/streetmerchant/issues/1035

This will be good to keep in mind when we start to move to the new config.

Thanks for the great contribution!

PR #1321, opened 2020-12-10
chore(store): add a few AMD listings for German stores

zeule:
### Description
RX 6800 and 6900xt  for amd-de and bits for mindfactory and alternate


jef (approved):
Great! Thanks for contributing 😄

PR #1336, opened 2020-12-11
feat(store): add darkhero motherboard to asus and microcenter

moodswung:
This adds support for the Asus Dark Hero motherboard.  It is currently scanning the Asus Store, Microcenter and NewEgg.


jef (commented):
Looks great!

I have two suggestions:

1. I think you'll want to use the model `crosshair viii`.
1. Do you mind updating the docs? You'll find it in `docs/filter/reference.md`

Other than that, looks great!

jef (approved):
Looks great! Thank you very much :shipit:

PR #1341, opened 2020-12-11
refactor: change `.env` to be `dotenv`

jef:
### Description

Refactored `.env` to be `dotenv` now. `.env` will still work until we release 4.0.0; then neither will work as it's a whole different system.

This is done because there are many cases where new users cannot find this file or have problems doing so. This should enable them to find and customize streetmerchant much easier.

### Testing

Local testing, seems to be working fine.


karlww:
I just updated to the latest version and .env does not work. My .env was being ignored completely until I renamed it to dotenv.

Issue #1394, opened 2020-12-14
Proxy - Cannot read property 'url' of null


MidnightsFX:
## Expected Behavior

No errors :)

## Current Behavior

[4:54:47 PM] info :: ℹ [amd] Next proxy index: 0 / Count: 2
[4:54:56 PM] error :: ✖ [bestbuy] evga 3080 ftw3 ultra - Navigation timeout of 30000 ms exceeded
[4:54:56 PM] info :: ℹ [bestbuy] Next proxy index: 1 / Count: 2
[4:54:57 PM] error :: ✖ [asus] asus 3080 strix oc - Navigation timeout of 30000 ms exceeded
[4:54:59 PM] error :: Cannot read property 'url' of null
[4:55:00 PM] error :: Cannot read property 'url' of null
[4:55:00 PM] error :: Cannot read property 'url' of null
[4:55:01 PM] info :: ✖ [bestbuy] [evga (3080)] ftw3 :: OUT OF STOCK

## Reproduction
1. Setup global.proxies
2. Run

## Environment
Windows 10 WSL2 - Ubuntu

```
# ** All configuration variables are optional **
######
USER_AGENT=
MAX_PRICE_SERIES_3080=850
MAX_PRICE_SERIES_RX6800XT=850
MAX_PRICE_SERIES_RX6900XT=1100
PAGE_BACKOFF_MIN=900000
PAGE_BACKOFF_MAX=1800000
PAGE_TIMEOUT=30000
SHOW_ONLY_SERIES=ryzen5600,rx6800xt,rx6900xt,3080
STORES=amazon:60000:90000,amd,asus,bandh:300000:600000,bestbuy,newegg:60000:90000,target,walmart,zotac
```

global.proxies
```
http://test-ip:8888
http://nsx-test-2:8888
```

Those DNS names are fake. I do have working ones in there.

I believe the issue here is that the proxies list is being accessed as zero indexed, so it is accessing proxies 0, 1, 2- however I only defined 2 proxies.

jef:
@Doridian, do you mind taking a look at this when you get the chance? Thanks!

Doridian:
The proxy list is being accessed correctly. As you can see it only says `Next proxy index: 0` and `Next proxy index: 1`.
However this `Cannot read property 'url' of null` error has happened to at least two different people now, and I want to figure out what it actually is. The only objects I know of that have a `url` property are stores and puppeteer request, and somehow the stack trace is missing, too...

[...]

PR #1408, opened 2020-12-14
fix(proxy): fix requests with proxies

Doridian:
### Description

As you might notice, I had to fork puppeteer-page-proxy. The fix I applied has already been PR'd upstream for several months and the library seems unmaintained currently (but there is no viable alternative either).
Other than that, I also added the `*.proxies` files to gitignore and fixed a problem where adblock would "randomly" disable request interception.

### DIFF
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -14,7 +14,8 @@ import {
 	delay,
 	getRandomUserAgent,
 	getSleepTime,
-	isStatusCodeInRange
+	isStatusCodeInRange,
+	noop
 } from '../util';
 import {disableBlockerInPage, enableBlockerInPage} from '../adblocker';
 import {config} from '../config';
@@ -23,7 +24,7 @@ import {filterStoreLink} from './filter';
 import open from 'open';
 import {processBackoffDelay} from './model/helpers/backoff';
 import {sendNotification} from '../notification';
-import useProxy from 'puppeteer-page-proxy';
+import useProxy from '@doridian/puppeteer-page-proxy';
 
 const inStock: Record<string, boolean> = {};
 
@@ -75,7 +76,7 @@ async function handleProxy(request: Request, proxy?: string) {
 	try {
 		await useProxy(request, proxy);
 	} catch (error: unknown) {
-		logger.error(error);
+		logger.error('handleProxy', error);
 		try {
 			await request.abort();
 		} catch {}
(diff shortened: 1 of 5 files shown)


Doridian:
Fixes https://github.com/jef/streetmerchant/issues/1394

src/store/lookup.ts
```diff
@@ -203,6 +204,11 @@ async function lookup(browser: Browser, store: Store) {
 						return onProxyFunc;
 					}
 
+					// Give dummy setRequestInterception to avoid AdBlock from messing with it
+					if (prop === 'setRequestInterception') {
+						return noop;
+					}
```
jef, line 210: I suppose the important part here is that you're returning an empty function rather than a value, correct?

Why is this important?

jef (approved):

src/store/lookup.ts
Doridian: The adblock plugin will sometimes, for various reasons, call `setRequestInterception(false)` because it thinks it is the authority on request listeners (which, due to us injecting proxies with that, too, it is not).
All this does is give the adblock plugin a function to call that will not error and simply does nothing at all, so it can be happy thinking it disabled request listeners, and the proxy plugin can be happy still using them.

jef: Aha, I see! Interesting. Do you think it would be worth mentioning making a ticket about on their repos?

Doridian: @jef Not really. The way puppeteer works, only one on-request listener is possible. The `Proxy` thing we do is a hack, and for sure not an expected use case of anything. Now if puppeteer ever supports multiple on-request listeners, then that might be considered an issue.

jef: Wait, I'm sorry. I'm misunderstanding the situtation. You did make a fork of package and you fixed that issue. As for https://github.com/cliqz-oss/adblocker/issues/1495, that isn't an issue for that library, yet an issue with the original `puppeteer-page-proxy`.

I'm I understanding this correct?

Doridian: @jef The issue with adblock purely arises because we need multiple things that employ request listeners, which puppeteer does not support in any way. So we hack around and "fake" the Page object the adblocker sees to be able to do it. Of course adblock has no way of expecting to get a proxied page object, so this is not a bug on their end at all, we just do something puppeteer cannot do.

PR #1411, opened 2020-12-14
feat: supporting (optional) custom playSound per store config

tmarshall:
### Description

While crawling sites for both Xbox & PS it can be useful to have a different sound per product. Doing that at the root (config) level is more complex, but doing it one-off per store config makes more sense. A separate store config per product can be added, with it's own `playSound` filepath.


src/notification/sound.ts
```diff
@@ -18,29 +19,33 @@ if (config.notifications.playSound) {
 	}
 }
 
-export function playSound() {
```
tmarshall, line 21: this file is much more readable if you view it w/o whitespace changes: https://github.com/jef/streetmerchant/pull/1411/files?diff=unified&w=1

PR #1422, opened 2020-12-15
fix(lookup): check out of stock before price

### LINKED ISSUE #1421: Some out of stock items are checked for max price before out of stock check
## Expected Behavior
Some out of stock items should be outputted as being out of stock by the console instead of showing that they are over their max price

## Current Behavior

Items that are out of stock and are over the set price limit are only shown as over the price limit

## Steps to Reproduce

e.g. Run the script with location "germany" and shop "proshop-de" and show_only_series "3070"

## Environment

**OS:**
Ubuntu Desktop 20.04
**dotenv file:**

as stated above

carrierfry:
Checking for Out of Stock before checking for max Price

### Description

Fixes #1421

### Testing

Tested it briefly on Ubuntu Desktop 20.04 and it worked

### New dependencies

none

### DIFF
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -347,6 +347,15 @@ async function lookupCardInStock(store: Store, page: Page, link: Link) {
 		}
 	}
 
+	if (store.labels.outOfStock) {
+		if (
+			await pageIncludesLabels(page, store.labels.outOfStock, baseOptions)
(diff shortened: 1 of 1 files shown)


jef (approved):
Ah, mhm. This makes sense to me!

Thanks for the contribution 😁

PR #1426, opened 2020-12-15
chore(store): add some radeon models for German stores

zeule:
### Description
add some Radeon models for German stores


jef (approved):
Great!

Thank you very much!

PR #1468, opened 2020-12-19
Update amazon-de.ts

KnotenJoe:
### Description

price checking was broken

### Testing
tested with different pricing

### DIFF
--- a/src/store/model/amazon-de.ts
+++ b/src/store/model/amazon-de.ts
@@ -16,7 +16,7 @@ export const AmazonDe: Store = {
 			text: ['in den einkaufswagen']
 		},
 		maxPrice: {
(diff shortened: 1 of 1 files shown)


Resaec:
I can not reproduce this container id on amazon.de pages. Can you provide a link to a page that uses this container?

selop:
I've checked some items which are on sale and some which aren't is still see `#priceblock_ourprice` as the only tag.

KnotenJoe:
I'm sorry. I was occupied over the holidays. I found the container id once for a 3080 model but it seems it was only a temporal change and was changed again. Sorry for wasting your time.

jef:
No worries! Thanks for the report 😀

PR #1479, opened 2020-12-20
feat(store): Add german stores equippr and futurex (RTX 3080 only)

Doerster:
### Description

Added german stores Equippr and Future-X.
Incl. RTX-3080 models to look up. Anything else one is looking for can be added.

(Not sure why it's acting like I thoroughly  replaced the whole index.ts - I just added the new stores)

### DIFF
--- a/src/store/model/equippr.ts
+++ b/src/store/model/equippr.ts
@@ -1,0 +1,176 @@
+import {Store} from './store';
+
+export const Equippr: Store = {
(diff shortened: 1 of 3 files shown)


jef:
Probably has to do with linting. I can fix it. Thanks!

jef:
Superseded by https://github.com/jef/streetmerchant/pull/1501

Thank you!

PR #1484, opened 2020-12-21
chore(newegg): add links for rx6800xt and rx6800

davidgauch:
…pphire rx6800 Pulse


jef (approved):
Great! Thank you very much!

PR #1493, opened 2020-12-22
Patch 1

DanRobinsonHPE:
Added 3 3060 Ti Cards

### Description

Added 3 3060 Ti Cards
There are more cards of other models that could be added, but I was only interested in these.

### Testing

Patched my own instance, bombed at first because I had xc instead of xc gaming.
Updated that and have it running now.

### New dependencies

None known.


jef (approved):

jef:
Awesome! Thank you :)

PR #1499, opened 2020-12-22
fix(util): proper cloudfare protection handling

### LINKED ISSUE #1490: Cloudfare protection handling does not work with top-user-agents
## Expected Behavior

Stores with Cloudfare protection should be able to wait and correctly load the webpage to capture the stock information.

## Current Behavior

Stores such as CCL (UK) do not work with the current implementation of DDoS protection handling due to the use of the `top-user-agents` module. It stays on the same product and shows as `CLOUDFARE, WAITING`

I did some digging and found that the reason was due to the use of unconventional user agents being utilised by the randomiser and therefore any Cloudfare stores do not progress onto the actual intended webpage.

## Steps to Reproduce

1. Scan for any products such as RTX 3060 Ti at CCL (UK) store
2. Notice that it will stay stuck on one card and show `CLOUDFARE, WAITING` endlessly

sohaiiil:
### Description

Fixes #1490 

- Use a different user agent randomiser with a desktop-only filter
- This prevents issue for any Cloudfare protected webpages (such as CCL or Caseking)

### Testing

1. Search for any products on a Cloudfare-protected store (e.g. CCL)
2. Notice that it works and doesn't stay in an endless loop as before

### New dependencies

- `user-agents`


neatchee:
Before merging this pull request, please confirm that the module you've used (user-agents) pulls ONLY from recent user-agents

A previous solution we were using was relying on a very stale user-agent list which was causing a significant number of problems on other sites. All of the modules we could find were using same or similar user-agent lists and we needed to manually set the top-50 useragents seen in the wild to resolve the issues.

@jef Should not be merged without first confirming the source of the user-agent list provided by this module.

neatchee:
Note: I have a separate pull request going up today that resolves the infinite "waiting" issue by aborting after 5 attempts. That will resolve issue #1490 albeit in a different way.

jef:
So I'm curious about the behavior before this? Beside the OS filter, all the UAs were desktop. (And even the most up to date, top user agents.) I'm not sure if I agree with filtering on Win32. What's the purpose behind that?

Thanks for the contribution! Just need some clarity 😊

jef:
> Before merging this pull request, please confirm that the module you've used (user-agents) pulls ONLY from recent user-agents

It seems that this pulls from top GA and CircleCI UAs, which would be good (basically the top). I haven't tested this in any way, so I'm hesitant to change again unless we have a problem.

Also, I think having other operating systems would be beneficial as well, not just Windows. Not certain though.

Thanks!

sohaiiil:
Im going to close this PR for now as it isn't relevant for the current context.

PR #1523, opened 2020-12-27
chore(philips-hue): correct the log text

collin-garvey:
### Description

Super minor change to the log output, it seems the text for these two blocks of the if/else were using the opposite verbiage

### Testing

Simply hooked up the Hue config and observed the log output locally


jef (approved):
Definitely could be misleading. Thanks!

PR #1548, opened 2020-12-29
chore(bestbuy): add 3060ti cards

Reid-Kornman:
removed duplicate cards and added new evga and pny cards


jef (approved):
Looks great. Thank you!

PR #1469, opened 2020-12-19
fix(galaxus): maxPrice selector

KnotenJoe:
Description
price checking was broken

Testing
tested with different pricing

PR #1568, opened 2020-12-31
Bugfix Galaxus, updated to .ZZf1 from old entry

wofnull:
Galaxus changed again the html class for the prices.

### Description
Update to #1469 from KnotenJoe 

Changed to  class .ZZf1 

### Testing

Galaxus gave false positives for AMD Princings, ignoring pricing at all. After patch pricing was checked / displayed correctly again.


wofnull:
Closing down Pull request as the change from galaxus happened again, and i patched it again in another pull request.

PR #1578, opened 2021-01-01
fix(galaxus): update selector

wofnull:
fix: Galaxus again changed the CSS marker for the product_details changed once before 2 days ago

### Description

Galaxus changed once more the CSS marker on the site, last time on 30.12. ( opened pull request #1568 , which as not processed in the meantime and got closed by me ) and beforehand as fixxed around a week ago from knotenjoe in #1469 

### Testing

Script  runs fine with the replaced CSS marker and pulls the correct price from the site again.

### DIFF
--- a/src/store/model/galaxus.ts
+++ b/src/store/model/galaxus.ts
@@ -8,7 +8,7 @@ export const Galaxus: Store = {
 			text: ['In den Warenkorb']
 		},
 		maxPrice: {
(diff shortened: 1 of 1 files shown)


jef:
This probably needs to get changed at some point to something else, as this will probably continue to be a problem. This is most likely because this selector is generated.

Thanks for keeping this updated in the mean time!

jef (approved):

PR #1590, opened 2021-01-03
feat: add command parameter to set custom dotenv conf

pbaum83:
Allows users to define the dovenv config file at run time. The reason for doing this was to allow running multiple instances with different configurations or to just maintain more than one configuration without having to replace the configuration file.

Example:
npm --conf=myenvUS start
npm --conf=myenvCA start

image
image

Note: I am not a JS pro and not sure this is the best way to do this. Also, I wasn't able to figure out another way to do this, if this is already default functionality please close.


pbaum83:
Fixed the build errors and rebased branch on origin/main.

jef (approved):
Thanks for doing this! Sorry it took awhile for me to pull it in.

# :rocket:

PR #1599, opened 2021-01-05
fix(galaxus): update selector

wofnull:
### Description

Follow up to #1578 

Galaxus changed again the CSS now to "Z1ex" leading  to the script not be able to pull the price and just the in stock properly

### Testing

Changed Variable, now works as intended again and catches Prices well


jef:
Thank you!

PR #1607, opened 2021-01-06
feat(galaxus): update selector, add out of stock check

wofnull:
### Description

Galaxus changed again the selector in the CSS, this time to Z1fp.

Last change done in: #1599 (yesterday)

### Testing

verified change, price is now tracked again and not just in stock / out of stock.


src/store/model/galaxus.ts
```diff
@@ -8,7 +8,7 @@ export const Galaxus: Store = {
 			text: ['In den Warenkorb']
 		},
 		maxPrice: {
-			container: '.productDetail .Z1ex',
+			container: '.productDetail .Z1f2',
```
jef, line 11: Alright, so after this being dynamic, I think we need another way to solve this as the updated selector no longer works either.

image

Let's see if we can come up with something else.

It seems that using just `.productDetail` works here. Do you want to test and update?

Thanks!

jef (changes requested):

src/store/model/galaxus.ts
wofnull: Using only ".productDetail"  works most parts, since i have some special trackings on the galaxus site aswell, i found that as soon that a number comes before the acutal price, the script takes that number (even if it s an percentage)

Example: https://www.galaxus.de/de/s1/product/lian-li-o11dynamic-xl-midi-tower-pc-gehaeuse-12200197
image
image
image
image

So it looks like it works until there is running something like a discount or something like this ( in this case )

jef: > i found that as soon that a number comes before the acutal price, the script takes that number (even if it s an percentage)

Well that sucks 😞

Also, updated selector again:

image

wofnull: Already updated like 10 minutes ago in bugfix branch, should be up to date now ( until the next change i think on sunday / monday ) ...

wofnull:
Out of the current not up to date selector another proble came up: all Items currently tracked were shown instock although the items were not instock, 
Currently the script checks for InStock via an selector if the cart is active or not, regardless of the response of this check, it seems the cart can be marked as active and be inactive ( or the check itself is not working as it should ), as workaround i have added  the 2 known states for out of stock on galaxus website as out of stock check.

Tested them and they work as expected , cross check with an real available item shows, instock is shown correctly at this stage.

However, the patch from 06.01 ( which  was not merged yet ) aswell as the updated patch from 08.01. ( which is aswell not merged yet ) is included. I suspect an additional change for the update marker today, but there is still no change visible yet. 

Please merge this asap, i will open another pull request if the state of the galaxus update marker changes again.

wofnull:
As suspected, the update selector changed again a few hours ago, fixed this with the latest patch ( together with the implementation of the out of stock check for galaxus  )

jef:
I guess we can continue to do this. Thanks for building out the outOfStock selectors.

jef (approved):

jef:
Thanks for all your help!

Issue #1610, opened 2021-01-06
Shorten best buy urls


kiowadriver:
## Expected Behavior

The most efficient means of cataloging and updating product URLs

## Current Behavior

We are using a more verbose url format than necessary in the bestbuy.ts store model. The BB api has a very very simple url that support both the product page and add cart function.

Examples:

Current URL in bestbuy.ts:
https://www.bestbuy.com/site/evga-geforce-rtx-3070-xc3-black-gaming-8gb-gddr6x-pci-express-4-0-graphics-card/6439300.p?skuId=.....

Shorter version using api:

Tracked Page:
https://api.bestbuy.com/click/-/6439300/pdp

Cart link:
https://api.bestbuy.com/click/-/6439300/cart

wofnull:
i have done a patch for this in #1613 , now we wait until jef approves / adds it to the working tree ;)

PR #1613, opened 2021-01-06
fix(bestbuy): shorter pageview urls and missing carturl links

wofnull:
### Description

Quickfix for #1610 + additional missing cartUrls in bestbuy.ts

- Updated all URLs to shorter format ( some were already in correct format , some where in other formats )
- added several carturls, that were missing in original file

### Testing

Verified that bestbuy runs through without issues on test script


jef (approved):
Oh awesome! This is pretty cool. Thanks for putting this together!

PR #1633, opened 2021-01-08
feat(proshop-de): add proshop-de inStock selector

benben:
### Description

fixes out of stock condition for proshop-de

### DIFF
--- a/src/store/model/proshop-de.ts
+++ b/src/store/model/proshop-de.ts
@@ -3,6 +3,10 @@ import {Store} from './store';
 export const ProshopDE: Store = {
 	currency: '€',
 	labels: {
+		inStock: {
+			container: 'button[data-form-action="addToBasket"]',
+			text: ['bestellen']
+		},
 		maxPrice: {
 			container:
 				'.site-currency-wrapper > span[class="site-currency-attention"]',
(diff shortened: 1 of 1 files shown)


jef:
Hey @benben, thanks for the contribution! I don't know if this actually changes anything as this text field gets moved to be all lowercase, so casing isn't a problem usually. Can you explain what's happening?

wofnull:
I have monitored issues with proshop today aswell and try to figure out the issue currently ( with not much success yet ).

As it seems Proshop-de.ts seems to report randomly change the standard out of stock text, but only for single views. Even if the browser is opened directly after from the bot, the update selector looks normal aswell as the search querry  for out of stock.

Currently every 10th to 15th out of stock message is therefore reporting as in stock. The problem came up first today and stopped for about 2 hours but came back again. Currently the behaviour is not present, could be aswell an issue with the site itself when its under heavy load ( was seeing white pages upon pageload when accessing product pages by hand where the problem came up first today )

jef:
Thanks for the report @wofnull. I think our best bet is to create an inStock property then, that way we don't get these false positives.

jef:
```typescript
		inStock: {
			container: 'button[data-form-action="addToBasket"]',
			text: ['bestellen']
		},
```

Can you test by adding this in?

wofnull:
Checked in #1635 as patch and verified working

benben:
Thanks for all you comments. I manually backported @wofnull's changes into this PR since the other PR includes a lot of unnecessary commits.

I tested in my own fork and it looks like it works now. It's running for >10 minutes now without false positives.

wofnull:
i closed my branch down now, was only attended for the case if this pull request gets somehow not updated anymore.

jef (approved):

PR #1635, opened 2021-01-08
fix(proshop-de): instock_check added

wofnull:
### Description

Patch for the test in #1633 

reverted the change from benben and added the instock feature. Changed aswell the testing model, since 30 series cards tend not to be available ( and therefore not so great for testing if the in stock works ) 
As mentioned in #1633 the proshop issue is randomly coming up and dissappearing, however the instock check should be more secure as only an outofstock check and should prevent on blank pages to trigger an instock message.

### Testing

after changing the test product i was able to check the functionality fully.


wofnull:
Closing pull request here, since beben already took over the needed patches in #1633

Issue #1637, opened 2021-01-08
Outdated/missing store.asus.com links


JesterEE:
## Expected Behavior
Query and report on stock of 30 series cards from store.asus.com

## Current Behavior
3060ti, 3070 cards are not present.  3080 and 3090 series card links are missing or outdated.

## Other
I created a feature branch to fix this, however I am not operating a development environment for this project and have no good way to test it for pull.  I request that @jef or one of the other project developers review the branch and pull/implement the updates I made to enhance streetmerchant.

JesterEE/streetmerchant@d04ba972c49e3e6f1371ccea9eee15b3a2b6a18c

wofnull:
you can open a pull request from your repo on github from within the branch with the regarding changes.

So jef can implement this directly, if the branch with the asus fix works on your end and is tested / runs through the github pretests, jef can directly import it on your behalf.

kiowadriver:
I tested the changes he made and they work if you also ensure you have added the "ko oc" to the store.ts file as well.

JesterEE:
@wofnull @kiowadriver
Thanks for helping with the testing!  I will submit it as a PR.

PR #1645, opened 2021-01-09
fix(store): Update 30 series links and models for store.asus.com

JesterEE:
### Description
Updates the links for 30 series cards on store.asus.com and adds the ASUS 'ko oc' as a Model variant.

### Testing
Testing complete by @wofnull and @kiowadriver in Issue #1637.


jef (approved):
Awesome! Thank you very much for updating :)

PR #1664, opened 2021-01-12
fix(galaxus): update selector  ( 14.01.2020 07:00 ) 

wofnull:
### Description

Next Galaxus Patch for changed update selector, this time its Z1cf

### Testing

tested and verified working in local installation.

### DIFF
--- a/src/store/model/galaxus.ts
+++ b/src/store/model/galaxus.ts
@@ -8,7 +8,7 @@ export const Galaxus: Store = {
 			text: ['In den Warenkorb']
 		},
 		maxPrice: {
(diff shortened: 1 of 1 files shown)


wofnull:
huh why is the build-and-lint failed on the twitch.ts ... i have not touched the file or changed it, the whole branch is forked from the main repo here ...

wofnull:
As it seems galaxus now changes the css selectors in their shop more often at least once every 24 hours down from 48 hours. Could be even more, since the galaxus part of the script runs only by hand 2-3 times a day from my end to check.

jef (approved):

Issue #1676, opened 2021-01-13
Price filter not working (maybe caused for numerical representation of the price)


liam-maps:
## Expected Behavior

The filter should work if the price is higher than the price I've set

## Current Behavior

The log shows a found product, even if the price is higher

## Steps to Reproduce

This product is over the thousands tag, but the thousands separator is a space, which I suspect is causing the issue
https://www.wipoid.com/asus-rog-strix-geforce-rtx-3080-oc-edition-10gb-gddr6x.html

With this also happens, but in this case the thousands separator is a dot, and the cents separator is a comma, because in Spain and other countries this is how it goes, which I suspect is causing the issue
https://www.vsgamers.es/product/tarjeta-grafica-asus-tuf-gaming-geforce-rtx-3090-oc-24g-24-gb-gddr6x

EDIT: Over thousand price triggered the same behaviour again, here:
https://www.amazon.es/dp/B08HM4V2DH

I don't know if this is the culprit or if you are taking into account all of the separator variants around the globe:
https://en.wikipedia.org/wiki/Decimal_separator#Examples_of_use

## Environment

Kubuntu 19.10

```
# ** All configuration variables are optional **
# Read https://github.com/jef/streetmerchant#customization for help on customizing this file
#############################################################################################

ASCII_BANNER=""
ASCII_COLOR=""
AUTO_ADD_TO_CART=""
BROWSER_TRUSTED=""
COUNTRY="spain"
DESKTOP_NOTIFICATIONS=""
DISCORD_NOTIFY_GROUP=""
DISCORD_WEB_HOOK=""
EMAIL_PASSWORD=""
EMAIL_TO=""
EMAIL_USERNAME=""
HEADLESS=""
IN_STOCK_WAIT_TIME=""
LOG_LEVEL=""
LOW_BANDWIDTH="true"
MAX_PRICE_SERIES_3060TI="550"
MAX_PRICE_SERIES_3070="650"
MAX_PRICE_SERIES_3080="750"
MAX_PRICE_SERIES_3090="1650"
MAX_PRICE_SERIES_CORSAIR_SF=""
MAX_PRICE_SERIES_RX6800=""
MAX_PRICE_SERIES_RX6800XT=""
MAX_PRICE_SERIES_RX6900XT=""
MAX_PRICE_SERIES_RYZEN5600=""
MAX_PRICE_SERIES_RYZEN5800=""
MAX_PRICE_SERIES_RYZEN5900=""
[...]

PR #1682, opened 2021-01-14
feat(vsgamers): add maxPrice selector

wofnull:
### Description

Fixes issue #1676 partly, for the vsgamers issue, since it was missing the maxPrice function in the config at all, giving false alarms with set price all over the place

### Testing
Testing done in test enviroment, everything seems to work as expected now.

image


jef (approved):
Awesome. Thank you!

PR #1698, opened 2021-01-15
fix(computeruniverse): false positives

wofnull:
### Description

Computeruniverse started listing most of the items "in Stock", basicly they added for 30 series cards another out of stock label.

added it to the out of stock list. 

### Testing

done and verified, items show now properly out of stock.


jef (approved):
Great! Thank you very much.

PR #1699, opened 2021-01-15
fix(acompc): false positives

wofnull:
### Description

AcomPC had showing every card as in Stock for the last few days, after checking I found that the out Of Stock text was now under the same selector as the in Stock one, but with another text.

Updated this to the correct selector

### Testing

Tested after changing out of stock is now working again as it should.


jef (approved):
Aha! Makes sense. Thanks for updating 🚀

PR #1714, opened 2021-01-17
fix(newegg): false positives (all stores)

wofnull:
### Description
During test run to lockout for new cards on newegg for another fix i do currently prepare, i found that the newegg site ( US / CA / SG ) does give false positives in streetmerchant, showing in Stock items but are in reality out of stock.

Checking against the config file, there was no out of stock check in place and only an instock check, which seemed to hit nothing when cards are not in stock, but still got triggered. Since all 3 sites use the same layout I added ( and tested ) a proper out of stock check.

### Testing
Testing has been done, Items out of stock show now proper out of stock and not the listed price / triggering false positives anymore.

### DIFF
--- a/src/store/model/newegg-ca.ts
+++ b/src/store/model/newegg-ca.ts
@@ -16,7 +16,17 @@ export const NeweggCa: Store = {
 			container:
 				'div#app div.product-price > ul > li.price-current > strong',
 			euroFormat: false
-		}
+		},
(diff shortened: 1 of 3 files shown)


jef (approved):
Good idea. I think this will help out a lot.

Thank you @wofnull for all the contributions lately! I really appreciate it.

PR #1743, opened 2021-01-19
chore(neweggca): add missing 3060ti, 3070, 3080, 3090 cards

wofnull:
### Description
Based upon #1741

Fittet all cards as before for newegg US / SG to local file requirements, removed the cart urls for this one, as they are missing completly and i do currently not have the actual links for them for the ca site. Furthermore, CA was somehow updated otherwise before and some cards from my US / SG update where already present, others still missing, but the old cards were all there.

this should sum up all updates for newegg for now

### Testing

Local test was running fine and gave the awaited cards back.


jef (approved):
Awesome! Thank you!

PR #1744, opened 2021-01-19
fix(notebooksbilliger): added 3060ti founders edition

wofnull:
### Description

Fixes #1729 

Added the missing FE card, checked for the link on Nvidia site for hidden page on notebooksbilliger. The provided mobile link uses the same ID, added the real link to notebooksbilliger.ts.

### Testing

Local test was running fine and gave the awaited card back.

### DIFF
--- a/src/store/model/notebooksbilliger.ts
+++ b/src/store/model/notebooksbilliger.ts
@@ -46,6 +46,13 @@ export const Notebooksbilliger: Store = {
         'https://www.notebooksbilliger.de/gainward+geforce+rtx+2070+super+phoenix+v1+grafikkarte+656238',
     },
     {
+      brand: 'nvidia',
+      model: 'founders edition',
+      series: '3060ti',
+      url:
+        'https://www.notebooksbilliger.de/nvidia+geforce+rtx+3060+ti+founders+edition+695677',
+    },
+    {
       brand: 'inno3d',
       model: 'ichill x3',
       series: '3070',


sohaiiil:
As stated before @wofnull, this won't work because the link changes every time a drop happens. Check the link now and it doesn't return any product (as expected).

wofnull:
@sohaiiil 
This works for notebooksbilliger.de just fine, as the link is hidden as soon as the cards go out of stock.
The link is reused for the next 3060ti drop as it is aswell with the 3080 FE / 3070 FE / 3090 FE on the same site.
Only thing that is currently not known is the second product link.

However you can verify that this is working for the 3060ti, if you check the mobile version of the site, where the product is not hidden beforehand:

https://m.notebooksbilliger.de/nvidia+geforce+rtx+3060+ti+founders+edition+695677

As already said, this should work, but can only be verified if a new 3060ti fe drop happens, but from the last 9 known drops of 3060ti cards on NBB the given link https://www.notebooksbilliger.de/nvidia+geforce+rtx+3060+ti+founders+edition+695677 was used as shop link

sohaiiil:
> @sohaiiil
> This works for notebooksbilliger.de just fine, as the link is hidden as soon as the cards go out of stock.
> The link is reused for the next 3060ti drop as it is aswell with the 3080 FE / 3070 FE / 3090 FE on the same site.
> Only thing that is currently not known is the second product link.
> 
> However you can verify that this is working for the 3060ti, if you check the mobile version of the site, where the product is not hidden beforehand:
> 
> https://m.notebooksbilliger.de/nvidia+geforce+rtx+3060+ti+founders+edition+695677
> 
> As already said, this should work, but can only be verified if a new 3060ti fe drop happens, but from the last 9 known drops of 3060ti cards on NBB the given link https://www.notebooksbilliger.de/nvidia+geforce+rtx+3060+ti+founders+edition+695677 was used as shop link

I believe this is still incorrect. There haven't been 9 drops of the RTX 3060 Ti on NBB. The 3060 Ti was only released on the 2nd of December. Since then, according to some online information I have found, there have been precisely 3 drops of the card on NBB.

- 02/12/20 - link unknown
- 17/12/20 - https://www.notebooksbilliger.de/nvidia+geforce+rtx+3060+ti+founders+edition+692948
- 13/01/21 - https://www.notebooksbilliger.de/nvidia+geforce+rtx+3060+ti+founders+edition+695677

This information is from _Stockdrops_ drop history table where they have recorded the links, dates and times for every drop of the FE card for all regions.

Based on this information, the link does in fact **change** every time. It does not stay the same.

jef:
We should probably not add this in since it changes so often.

wofnull:
the main problem i see here is that we not clearly now how the drop system / market system of nbb clearly works.

We have for once the drop based SKUs that are appearing when an 3060ti / 3070 / 3080 / 3090 drop happens, and got blanked after they are out  of stock again.
And then we have aswell the SKUs on NBB that have been around since the beginning, and still have the site content after the drops ( and showing stock during the stocks ) . Problem here on hand is that we have only the ones for 3090 / 3080 / 3060ti ( the link @sohaiiil ) provided.

a) the FE links in the notebooksbilliger.ts file for 3090/3080/3070 are broken and will not work at all 
b) a few of them are simply wrong

but wihout an actual drop when the site is currently monitored and the links are verified its only an assumption ...

sohaiiil:
> We should probably not add this in since it changes so often.

Yeah this can be closed as this doesn't resolve anything 👍

Issue #1753, opened 2021-01-20
Parsing bug! Not detecting in-store products as in-stock for Micro Center


hifihedgehog:
I am trying to help my fellow users at my local Micro Center in Madison Heights, MI (storeid=055) to see stock levels now that the store management has banned staying on the premises during afterhours. The problem is streetmerchant is not detecting in stock items.

As a test, I added the SF 600 Gold, an item I know is in stock, [following the directions here.](https://jef.codes/streetmerchant/help/general/)

[Clearly, I see stock.](https://www.microcenter.com/product/461588/corsair-sf600-600-watt-80-plus-gold-sfx-fully-modular-power-supply?storeid=055) However, streetmerchant does not and therefore sends no alerts out.

Here are the steps in brief to reproduce the problem.

I edited store.ts to include "600 gold" as a model. I did not added series or brand since Corsair and SF were already present under their respective export types. config.ts did not require editing since it only pertains to series.

Finally, I edited .dotenv to include my Micro Center location (MICROCENTER_LOCATION=madison-heights), the product series SF so it shows SF Corsair power supplies (SHOW_ONLY_SERIES=sf), and, of course, Micro Center as a store (STORES=microcenter).

Here is the weird thing. The console shows it is scanning the right products (SF600 and SF750), but it is not showing the SF600 as in stock....

![image](https://user-images.githubusercontent.com/16614343/105142223-56229c80-5ac8-11eb-95f6-72370b3298ec.png)

...when if I visit the page in my browser on the same machine, it most certainly is.

![image](https://user-images.githubusercontent.com/16614343/105142354-7eaa9680-5ac8-11eb-9be4-8425e0affb79.png)

Thoughts?

hifihedgehog:
As a gut check, I added a price limit of $100 (MAX_PRICE_SERIES_CORSAIR_SF=100). It is certainly parsing that correctly:

![image](https://user-images.githubusercontent.com/16614343/105144064-af8bcb00-5aca-11eb-907c-691f4a0da755.png)

hifihedgehog:
There is a parsing bug! I definitely don't see any occurrences of "in stock" under #cart-options under view-source:https://www.microcenter.com/product/461588/corsair-sf600-600-watt-80-plus-gold-sfx-fully-modular-power-supply?storeid=055. However, I do see it under #pnlInventory. Change the following in microcenter.ts to read like so:

```
export const MicroCenter: Store = {
  currency: '$',
  labels: {
    inStock: {
[...]

PR #1754, opened 2021-01-20
fix(microcenter): properly flag in stock items

hifihedgehog:
Fixes issue #1753


jef (approved):
Great! Thank you very much!

PR #1814, opened 2021-01-27
fix(alternate): update URLs

frommie:
Fixed the URLs for alternate (de), deleted duplicate entry for PS5 Disc Edition.

### Description

Fixes status error code 404 for alternate

### Testing

Alternate now returns correct stock status


frommie:
As I see the deleted entry is the bundle so maybe we should leave it

jef (approved):
Great! Thanks for fixing ☺️

PR #1821, opened 2021-01-28
add turkey stores

fastengineer:
Added stores in Turkey


fastengineer (commented):
Looks good

wofnull:
i looked through your update:
you added a bunch of additional config parts duplicating most of the existing items, this is uneccessary and will lead to confusion for user.
Additionaly you added last gen stuff ( 3000 Series Ryzen / 5700xt ) which is aswell not meant to be added in the main branch.

Further the built and lint failed in over 2000 Codeparts of your update , this needs to be fixxed before anything at all can be added.

fastengineer:
> i looked through your update:
> you added a bunch of additional config parts duplicating most of the existing items, this is uneccessary and will lead to confusion for user.
> Additionaly you added last gen stuff ( 3000 Series Ryzen / 5700xt ) which is aswell not meant to be added in the main branch.
> 
> Further the built and lint failed in over 2000 Codeparts of your update , this needs to be fixxed before anything at all can be added.

Yes I duplicated most of them, because alarm currency is different for them. We need alarm price definition with currency. 
Also I couldn't understand why there is no database for the items and item attributes. All data is hardcoded.

wofnull:
> Yes I duplicated most of them, because alarm currency is different for them. We need alarm price definition with currency.
> Also I couldn't understand why there is no database for the items and item attributes. All data is hardcoded.

Since most stores are regional only, and do not ship to other countrys ( or currency regions ) this is indeed not needed to be separated. The Setting for US / EU Payment variant is mostly coded into the <store>.ts files , but are limited to only define if the 1000 separator sign is a , or a . 

The database thing needs to be explained by @jef

lattedesu:
I would love to see this implemented!

rfve:
@fastengineer I want to make notification settings for telegram or pushbullet. which file I need to change. 
2- I just amazon.de amazon.it to scan the  and turkish stores. which file do I need to edit for this?

@wofnull @joergbattermann @jef

PR #1827, opened 2021-01-28
fix: Updating Irish Stores

SavageOranges:
Update to existing Irish store:

*Enables PS5 Digital Edition checks

### Description

Smyths Ireland only searching for disc edition console, this adds support for digital edition.

### Testing

Checked stock on my own client of both digital and disc editions. Both listings displayed in my client.

Replace smyths-ie model file with this one and be searching for disc and digital edition PS5 stock.


jef (approved):
Thank you! Looks good. Just need to fix the CRLF issues!

PR #1836, opened 2021-01-29
feat(store): add mediamarkt austria

mreilaender:
### Description

Adds support for [mediamarkt.at](https://www.mediamarkt.at/)

### Testing

- Included added store to the configuration, replaced a link from a graphics card with a product that exists and is available
- started the application and the graphic card model of the replaced link was shown in the logs

### DIFF
--- a/src/store/model/mediamarkt-at.ts
+++ b/src/store/model/mediamarkt-at.ts
@@ -46,44 +46,51 @@ export const MediamarktAt: Store = {
       brand: 'gainward',
       model: 'phoenix',
       series: '3060ti',
-      url: 'https://www.mediamarkt.at/de/product/_gainward-grafikkarte-geforce-rtx-3060-ti-phoenix-golden-sample-8gb-ga-3060rtx-ti-ph-gs-1815563.html',
+      url:
+        'https://www.mediamarkt.at/de/product/_gainward-grafikkarte-geforce-rtx-3060-ti-phoenix-golden-sample-8gb-ga-3060rtx-ti-ph-gs-1815563.html',
     },
     {
       brand: 'gainward',
       model: 'phantom gaming',
       series: '3080',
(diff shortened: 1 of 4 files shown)


jef (dismissed):
Looks good to me! Thank you for the contribution! 💯

src/store/model/mediamarkt-at.ts
```diff
@@ -0,0 +1,96 @@
...
+      },
+    ],
+  },
+  links: [
+    {
+      brand: 'test:brand',
+      model: 'test:model',
+      series: 'test:series',
+      url: 'https://www.mediamarkt.de/de/product/-2641856.html',
+    },
```
jef, line 44: I would change this to use mediamarkt.at as well if you don't mind.

src/store/model/mediamarkt-at.ts
```diff
@@ -0,0 +1,96 @@
...
+    {
+      brand: 'gainward',
+      model: 'phoenix',
+      series: '3060ti',
+      url:
+        'https://www.mediamarkt.at/de/product/_gainward-grafikkarte-geforce-rtx-3060-ti-phoenix-golden-sample-8gb-ga-3060rtx-ti-ph-gs-1815563.html',
```
jef, line 50:
```suggestion
        'https://www.mediamarkt.at/de/product/-1815563.html',
```

Also seems we can normalize these urls by using this. Do you mind changing that?

mreilaender: Done

jef (commented):
Awesome! Everything looks great. Do you mind adding this into the documentation?

Thank you!

mreilaender:
@jef The list of supported stores is all I have to add to the doc right?

Also thanks for your awesome work. Your code looks absolutely gorgeous :-)

jef (approved):
Great! Thank you!

PR #1870, opened 2021-02-03
fix: mis-placed 3070 Vision OC

aguilaair:
### Description

Right now there is a 3070 which is being registered as a 3080 eagle OC, this fixes the series and the model of the graphics card on the wipoid store

image


aguilaair:
@wofnull fixed some other stuff on https://github.com/jef/streetmerchant/pull/1879, so I'm going to close this

PR #1872, opened 2021-02-03
feat(notification): add Streamlabs support

TomZanna:
### Description

Added the ability to activate alerts via StreamLabs

Requires the user to register with StreamLabs and create an application to request an access token (no need to submit for review).
Useful for sending sound notifications to viewers of a live stream.

### Testing
Ran the test notification script

### New dependencies
form-data


TomZanna:
Over the weekend I will fix the linting problems.
Do you think I could avoid using `form-data`? I don't like that part of code but it just works.

jef (approved):
Query params kind of make sense, but we also didn't make the API! That being said, I feel like authentication shouldn't be sent as a query param, but unfortunately, it is.

Overall, looks good to me!

Thanks for the implementation 😄

docs/reference/notification.md
```diff
@@ -209,3 +209,16 @@ Instructions on how to set up tokens can be found at [d-fischer.github.io/twitch
...
+Instructions on how to set up tokens can be found at [dev.streamlabs.com](https://dev.streamlabs.com/docs/register-your-application).
+You don't need to submit your application for review, just whitelist yourself!
+
+| Environment variable | Description |
+|:---:|---|
+| `STREAMLABS_ACCESS_TOKEN` | StreamLabs access token |
+| `STREAMLABS_TYPE` | StreamLabs alert type |
+| `STREAMLABS_IMAGE`| Custom image to display. Leave it blank for default |
+| `STREAMLABS_SOUND` | Custom image to play. Leave it blank for default |
+| `STREAMLABS_DURATION` | StreamLabs alert duration |
```
jef, line 224:
```suggestion
| `STREAMLABS_DURATION` | StreamLabs alert duration (in milliseconds) |
```

src/config.ts
```diff
@@ -332,6 +332,13 @@ const notifications = {
     consumerSecret: envOrString(process.env.TWITTER_CONSUMER_SECRET),
     tweetTags: envOrString(process.env.TWITTER_TWEET_TAGS),
   },
+  streamlabs: {
+    accessToken: envOrString(process.env.STREAMLABS_ACCESS_TOKEN),
+    type: envOrString(process.env.STREAMLABS_TYPE),
+    imageHref: envOrString(process.env.STREAMLABS_IMAGE),
+    soundHref: envOrString(process.env.STREAMLABS_SOUND),
+    duration: envOrNumber(process.env.STREAMLABS_DURATION),
+  },
```
jef, line 341:
```suggestion
  streamlabs: {
    accessToken: envOrString(process.env.STREAMLABS_ACCESS_TOKEN),
    duration: envOrNumber(process.env.STREAMLABS_DURATION),
    imageHref: envOrString(process.env.STREAMLABS_IMAGE),
    soundHref: envOrString(process.env.STREAMLABS_SOUND),
    text: {
        header: envOrString(process.env.STREAMLABS_HEADER),
        subheader: envOrString(process.env.STREAMLABS_SUBHEADER),
    },
    type: envOrString(process.env.STREAMLABS_TYPE),
  },
```

I'm wondering if we should take advantage of the `message` and `user_message` params?

src/notification/streamlabs.ts
```diff
@@ -0,0 +1,42 @@
+import {Link, Store} from '../store/model';
+import {Print, logger} from '../logger';
+import {config} from '../config';
+import {URLSearchParams} from 'url';
+import fetch from 'node-fetch';
+
+const {streamlabs} = config.notifications;
+let requestParams: URLSearchParams;
+
+if (streamlabs.accessToken) {
```
jef, line 10:
```suggestion
if (streamlabs.accessToken && streamlabs.type) {
```

Since `type` is required, I think we should put this here as well.

src/notification/streamlabs.ts
```diff
@@ -0,0 +1,42 @@
...
+  requestParams = new URLSearchParams();
+  requestParams.append('access_token', streamlabs.accessToken);
+  requestParams.append('type', streamlabs.type);
+  requestParams.append('image_href', streamlabs.imageHref);
+  requestParams.append('sound_href', streamlabs.soundHref);
+  requestParams.append('duration', streamlabs.duration.toString());
```
jef, line 16:
```suggestion
  requestParams.append('access_token', streamlabs.accessToken);
  requestParams.append('duration', streamlabs.duration.toString());
  requestParams.append('image_href', streamlabs.imageHref);
  requestParams.append('message', streamlabs.text.header);
  requestParams.append('sound_href', streamlabs.soundHref);
  requestParams.append('type', streamlabs.type);
  requestParams.append('user_message', streamlabs.text.subheader);
```

Maybe from previous suggestion?

jef (commented):
Okay, just a couple comments. This is great overall; awesome addition!

src/config.ts
TomZanna: yeah, we could but as it for now the message parameter will be overwritten by the output of `Print.inStock (...)`.

I am thinking of two options:

1. View the product in the header and the store in the subheader
2. Display both the product and the store in the header and a custom message in the subheader (or vice versa)

jef: Oh! Right... Sorry, I totally forgot about that line. Okay, that sounds great. Let's just not change this for now and if we want to implement at some point, then we can. It's not a great improvement, so this is good to me!

Thanks for the help ❤️

jef (approved):

jef (approved):

PR #1880, opened 2021-02-04
feat: add restart time for leaking Chromium in Ubuntu

Fmstrat:
### Description

In Ubuntu, when attempting to check frequently and as Chromium tabs grow and usage increases over time, memory management suffers and machines will lock up (tested across multiple PCs, and happens in other heavy-use Puppeteer applications as well). While `nice`ing helps, it does not fully solve the problem.

This PR adds a `RESTART_TIME` flag that takes an integer in `ms` to set when streetmerchant will restart Chrome, thus resetting memory usage.

### Testing

- Set up a machine with 4-8GB RAM running Ubuntu 20.04
- Use streetmerchant as default checking all sites
- Let run overnight
- Wait for lockup or reboot

**NOTE:** Wiki updates required after PR. Happy to do this as well.


Fmstrat:
I realized after the fact that the documentation was actually in the repo, so I've added that to the PR.

jef (approved):
This makes a lot of sense to have in the meantime. I really appreciate your help here! :rocket:

jef (approved):

Issue #1884, opened 2021-02-04
1851 breaks the code for amazon-ca


augustynr:
## Expected Behavior

The patch 1851 brakes the code for amazon-ca
## Current Behavior
lots of errors

## Steps to Reproduce
the clean new pull of the source

## Environment

- OS:
ubuntu 20.04

```dotenv
```
STORES="amazon-ca"

more
## Logs

[11:32:26 a.m.] info :: ℹ selected stores: amazon-ca
[11:32:26 a.m.] info :: ℹ selected series: 3060ti, 3080
[11:32:26 a.m.] info :: ✔ sound player found: aplay
[11:33:40 a.m.] info :: ✖ [amazon-ca] [msi (3080)] gaming x trio :: OUT OF STOCK
[11:33:47 a.m.] error :: ✖ [amazon-ca] msi 3080 ventus 3x oc - Cannot read property 'join' of null
[11:33:54 a.m.] error :: ✖ [amazon-ca] evga 3080 ftw3 - Cannot read property 'join' of null

wofnull:
checked against the current release, as it seems we have here the same issue as in previous fixxes  where i found that the outOfStock check was somehow missing and messing up the results. i will do a quickfix for this.

PR #1851, opened 2021-02-01
fix(amazon-ca): update selector 

wofnull:
### Description
for maxprice changed for the whole store

Fixes issue #1846 

### Testing

Local Testing done and verified via CA VPN:

image

PR #1885, opened 2021-02-04
fix(amazon-ca): out of stock check was missing

wofnull:
### Description

Fixes: #1884

During the testrun of #1851 the errors where not triggered. but i think somehow that the missing outofstockcheck in any store will cause problems at some point.

### Testing

Testing done via VPN ( CA ) and directly to find additional issues, after adding out of stock check everything worked fine here.
image


jef (approved):
Awesome! Thank you very much @wofnull 🚀 ☺️

PR #1893, opened 2021-02-05
Workaround Amazon FR - PS5 standard - Alert incorrect

SchwarzDeveloping:


wazerstar:
Try run headless=false, you will see any does not work and does not select any of the ps5's

PR #1962, opened 2021-02-16
chore(pccomponentes): add 3090, 3060ti, rx6900xt

urshulgi:
Added all the 3090, 3060ti, and rx6900xt series from pccomponentes.

### Description

Pccomponentes was only searching for 3070 and 3080 series. This adds other series currently available on the website.

### Testing
Checked current stock and as there is none it shows out of stock. If necessary I will do further testing.


jef (changes requested):
Great! Thank you!!!

jef (approved):

PR #1989, opened 2021-02-18
feat(store): add netonnet-no

davols:
### Description
Added another norwegian store, netonnet.no - Since it exists in Sweden I let it have -no suffix. 

### Testing
I've tested it with an in stock 3090 which checks out as in-stock and the obvious out of stock 3080's and 3060ti's.


jef:
> Since it exists in Sweden I let it have -no suffix.

I promise I don't have a bias towards America and other countries, it just happened this way 😆

Thank you very much for the contribution!

jef (approved):

PR #2003, opened 2021-02-21
chore(pcdiga): add 3090, 3070, 3060ti

jcalado:
### Description

Initial iteration of the pcdiga website support only added 3080 series.
This should improve and expand to support also the 3090, 3070 and  3060ti series.

### Testing
Everything is out of stock for now, but tested with dotenv STORES=pcdiga


jef (approved):
Great! Thank you very much.

PR #2004, opened 2021-02-21
feat(store): add globaldata

jcalado:
### Description

Adds support for a major PC retailer in Portugal, Globaldata

### Testing

Tested locally - stock status working.
Can't get maxPrice since the element has `&nbsp;` to separate thousands, breaking the regex I guess..

### DIFF
--- a/src/store/model/globaldata.ts
+++ b/src/store/model/globaldata.ts
@@ -20,7 +20,7 @@ export const Globaldata: Store = {
       model: 'test:model',
       series: 'test:series',
       url:
-        'https://www.globaldata.pt/grafica-asus-geforce-rtx-3080-ek-10gd6x-90yv0f60-m0nm00',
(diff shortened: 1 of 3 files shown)


src/store/model/globaldata.ts
```diff
@@ -0,0 +1,329 @@
...
+      text: ['Notifique-me quando estiver em stock'],
+    },
+  },
+  links: [
+    {
+      brand: 'test:brand',
+      model: 'test:model',
+      series: 'test:series',
+      url:
+        'https://www.globaldata.pt/grafica-asus-geforce-rtx-3080-ek-10gd6x-90yv0f60-m0nm00',
```
jef, line 23:
```suggestion
        'https://www.globaldata.pt/drive-externa-dvdr-asus-sdrw-08d2s-u-lite-8x-branco-usb-20-90-dq0436-ua221kz',
```

Test model should have something that is usually in stock.

jef (commented):
Other than the one suggestion, everything else looks great!

Thank you 💯

jef (approved):
Great! Thank you!

PR #2021, opened 2021-02-24
chore(store): add some PS5 bundles

DiestelmannRaffael:
Some German stores now favor bundles over the plain consoles. With this commit I added some of these bundles to the correspondent store models


erwinc1:
@DiestelmannRaffael please use spaces instead of tabs

jef (approved):

PR #2029, opened 2021-02-25
chore(store): add 3060 products

JW6677:
### Description

Added 3060 support for street merchant and I've also added the required config lines to dotenv-example.
I have also added the 3060 links to some of the UK stores and one European one. Unfortunately at the moment, I can't add every single link for every single store for every single country.

### Testing

I've run the bot with the 3060 series selected for over 8 hours without any issues. All is working well.


jef (approved):
Awesome! Thank you!

PR #2034, opened 2021-02-26
feat: Refactor notifications and make them injectable

gmalette:
### Description

First I think it's worth giving context to what my goal is. I would like to create a program that wraps streetmerchant and offers another dashboard. I would like for it to run in the same process and avoid adding infrastructure if at all possible.

This PR refactors the notifications to make both failures and success notifications. It also makes the notification function injectable, but the `main` function uses the existing one.

I also refactored to make the logging of failures as part of normal notifications. This isn't strictly necessary but I thought it was nice to harmonize both and show that logging is just one way to handle poll events.

This PR, along with https://github.com/jef/streetmerchant/compare/main...gmalette:gm/packaging?expand=1, allow me to use streetmerchant and inject notification handling.

```ts
import {Browser, launch} from 'puppeteer';
import {config} from 'streetmerchant/build/src/config';
import {getSleepTime} from 'streetmerchant/build/src/util';
import {storeList} from 'streetmerchant/build/src/store/model';
import {tryLookupAndLoop} from 'streetmerchant/build/src/store';
import {LinkPollEvent} from 'streetmerchant/build/src/notification/link_poll_event';

const sendNotification = (event: LinkPollEvent): void => {
  console.log(event);
};

async function main() {
  const args: string[] = [];
  const browser: Browser = await launch({
    args,
    defaultViewport: {
      height: config.page.height,
      width: config.page.width,
    },
    headless: config.browser.isHeadless,
  });

  config.browser.userAgent = await browser.userAgent();

  for (const store of storeList.values()) {
    if (store.setupAction !== undefined) {
      store.setupAction(browser);
    }

    setTimeout(tryLookupAndLoop, getSleepTime(store), browser, store, sendNotification);
  }
}

void main();
```

### Testing

Ran the program, worked as normal for "out of stock" and other errors.
I was not able to test for "in stock" items.

PR #2038, opened 2021-02-26
feat: allow using streetmerchant as a library

gmalette:
### Description

As explained in #2034, I'm trying to use streetmerchant as a library in another project that exposes a different dashboard.
To allow this, I need to make a few changes to this codebase. This PR contains 3 changes:

1. It adds a `version` number to `package.json`, as `npm` requires a version to be present.
2. It loads the `version.txt` file from a path relative to `banner.ts`.
3. It bumps the support to node `<15.11.0`. This is not necessary.

### Testing

Ran the program, still works


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "streetmerchant",
+  "version": "3.4.1",
```
gmalette, line 3: Adding a version number here will increase the maintenance burden because it requires updating versions in two places. I can change `banner.ts` to use this number if that's preferable. 

The changes in this file aren't strictly necessary as it's easy to make them myself and then tell `git` to make them `assume-unchanged`.

package.json
```diff
@@ -17,7 +18,7 @@
     "test:notification:production": "node build/test/functional/test-notification.js"
   },
   "engines": {
-    "node": ">=12.0.0 <15.0.0"
+    "node": ">=12.0.0 <15.11.0"
```
gmalette, line 21: I understand that there could be good reasons for keeping `<15.0.0`, which I don't know.

src/banner.ts
```diff
@@ -1,7 +1,8 @@
 import chalk from 'chalk';
 import {readFileSync} from 'fs';
+import path from 'path';
 
-const version = readFileSync('version.txt', 'utf8');
+const version = readFileSync(path.join(__dirname, '../../version.txt'), 'utf8');
```
gmalette, line 5: I would really like to get this change in if possible. I'm not terribly familiar with the typescript compiler nor the node runner, but when I run it from my other project, the `version.txt` file isn't found as it's looking for it somewhere else.

package.json
jef: I was keep it below 15 only because some folks were having problems with `tsc`. We should probably get on latest sooner than later though!

package.json
jef: Before we make this change, I'd like to update the pipeline to manage this for us. This is easy enough, but just needs an update.

src/banner.ts
jef: Interesting... Yeah I'm getting the version running as is, but if you're running within Docker, this could be an issue?

jef:
Alright, made a PR (#2066)

This should get the things you need!

src/banner.ts
gmalette: I think it was looking in the path of the project that wraps it.

gmalette:
Alright! Thanks a bunch!

PR #2039, opened 2021-02-26
chore(newegg): add 3060 links

SalamUchiha:
I just updated the newegg.ts file with the 3060 links, also added in the names for the new models in the store.ts file


jef (approved):
Great, thank you!

PR #2041, opened 2021-02-26
chore(store): add 3060 for mediamarkt, saturn and notebooksbilliger

frzme:
### Description

Adds all RTX3060 I could find to notebooksbilliger and mediamarkt/saturn
For mediamarkt/saturn I could only find 2 product URLs for the time being.

### Testing

I ran the bot with the added URL and it seems to work fine.
I verified all URLs to work.


jef (approved):
Great addition. Thank you!

PR #2043, opened 2021-02-27
feat(store): add novoatalho (PT)

afonsojramos:
### Description

Added Novo Atalho, a popular Portuguese store.

Also sorted some of the entries. OCD was kicking in 😅

### Testing

Tested locally.
Similar to #2004 can't get maxPrice since the element has &nbsp.


jef (approved):
Haha, I don't mind the OCD at all 💯 

I appreciate it and thanks for updating those docs!

PR #2053, opened 2021-02-28
feat(store): add laptopsdirect (IE) 

BeyhanSen:
### Description
Added laptopsdirect Ireland.

### Testing
Tested locally.


src/store/model/laptopsdirect-ie.ts
```diff
@@ -0,0 +1,24 @@
+import {Store} from './store';
+
+export const LaptopsdirectIE: Store = {
+  currency: '€',
+  labels: {
+    inStock: {
+      container: 'span.atb',
+      text: ['Add to Basket »'],
```
jef, line 8:
```suggestion
      text: ['Add to Basket'],
```

This should be sufficient

jef (commented):
For me to merge this in, I would like to see a test model as well as out of stock label if you don't mind!

jef (commented):
For me to merge this in, I would like to see a test model as well as out of stock label if you don't mind!

jef (changes requested):
For me to merge this in, I would like to see a test model as well as out of stock label if you don't mind!

PR #2057, opened 2021-03-01
fix(lookup): check in stock before price

Quentin-M:
### Description

I long meant to make that simple change.. In the same vein as #1422, ensure the item is in stock before printing an error based on a price check. This makes the output of streetmerchant more useful as out of stock items will appear out of stock.. and only when an item is in stock but does not match our price requirements, will we print the price range mismatch. If the price is just barely out of the range but seems still acceptable for the user, it gives them the actionable opportunity to place an order. Otherwise, the price mismatch messages are simply noise.

### DIFF
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -438,17 +438,6 @@ async function lookupCardInStock(store: Store, page: Page, link: Link) {
     }
   }
 
-  if (store.labels.maxPrice) {
-    const maxPrice = config.store.maxPrice.series[link.series];
-
-    link.price = await getPrice(page, store.labels.maxPrice, baseOptions);
-
-    if (link.price && link.price > maxPrice && maxPrice > 0) {
-      logger.info(Print.maxPrice(link, store, maxPrice, true));
-      return false;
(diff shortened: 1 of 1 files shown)


jef:
This is interesting, because we have two lines of thought:

- Basically what you said, the user can determine if they want to buy
- The user puts a max price and that _is_ their max price.

In some cases, the price may be adjusted by 10-20 bucks. But it most cases it's a few hundred.

I would be more interested in implementing a range, rather than switching these around, as the maxPrice check would never get ran if we change this.

marvelmaster:
is this implemented yet? because latest doesnt come with this.

jef:
I believe the intent is good, but not what we want _exactly_.

**Currently implemented**

- Regardless if the item is in or out of stock, if it's above our price point, tell the user.

**In this PR:**

- If the item is in stock, tell the user.

**Proposed solution:**

- If the item is in stock, check if it's above our max price, if not, tell the user.

---

Seems like @Quentin-M deleted their fork/branch, so I will continue this effort on another PR.

PR #2072, opened 2021-03-03
feat(store): add amd-nl store

thecodeassassin:
Added AMD-NL store


bramevo:
I've added BE store. So might want to add/review both at the same time. #2338 
I did notice that the site redirects based on your location so might be hard to review them. In my case it's just a copy paste of the de version with language and location update, that's it.

jef:
> I did notice that the site redirects based on your location so might be hard to review them.

Okay, this is very interesting... So you're saying if you went to https://www.amd.com/en/direct-buy/5450881400/us it redirects you to https://www.amd.com/en/direct-buy/5450881400/be? (Assuming you're in Belgium.)

> In my case it's just a copy paste of the de version with language and location update, that's it.

A lot of the stores are :laughing: Perhaps we could refactor this.

bramevo:
> > I did notice that the site redirects based on your location so might be hard to review them.
> 
> Okay, this is very interesting... So you're saying if you went to https://www.amd.com/en/direct-buy/5450881400/us it redirects you to https://www.amd.com/en/direct-buy/5450881400/be? (Assuming you're in Belgium.)

Correct, redirects to https://www.amd.com/en/direct-buy/5450881400/be
Not always the same though. For example, ID is different in the us for the RX 6700 XT. Our link is https://www.amd.com/en/direct-buy/5496921400/be . Also we have a midnight version (limited edition) of the 6800 XT card.

> 
> A lot of the stores are 😆 Perhaps we could refactor this.
> 
> Edit: Interesting, going to https://www.amd.com/en/direct-buy/5450881400 vs `/us` vs `/be`, all go to the same URL. I'd be interested to remove all countries and just use `amd` -- although the selector text could change. Is it all in English?

So not always the same urls. Language could be the same, although https://www.amd.com/fr/direct-buy/5496921400/be is the french version which also exists.

jef:
Well, in that case, I don't want to over complicate it to reduce a couple files -- considering we have over 100. This will be part of a larger effort of refactoring that'll come in 4.0.0.

Thanks for your insight!

jef (approved):

PR #2089, opened 2021-03-04
feat(store): Add AO and BT (UK)

ruippeixotog:
### Description

Adds support for PS5 on AO and BT. In both cases searching for PS5 takes users to a generic landing page, so I'm targeting the messages shown there.

### Testing

Tested locally and ran them for some time in my own quest for a PS5.


jef (approved):
Works with me! Thank you!

PR #2094, opened 2021-03-05
chore(amd): update rx6900xt urls

poulet42:
### Description

6900RX URL have changed, it is now [https://www.amd.com/en/direct-buy/5458374200/[locale]](https://www.amd.com/en/direct-buy/5458374200/us)

### Testing

If you go to the former [URL](https://www.amd.com/en/direct-buy/5458372200/us). This will redirect you to the list of products, and wrongly state that the product is in stock, changing to the new url fixes that issue

### DIFF
--- a/src/store/model/amd.ts
+++ b/src/store/model/amd.ts
@@ -54,10 +54,10 @@ export const Amd: Store = {
     {
       brand: 'amd',
       cartUrl:
-        'https://www.amd.com/en/direct-buy/5458372200/us?add-to-cart=true',
+        'https://www.amd.com/en/direct-buy/5458374200/us?add-to-cart=true',
       model: 'amd reference',
       series: 'rx6900xt',
-      url: 'https://www.amd.com/en/direct-buy/5458372200/us',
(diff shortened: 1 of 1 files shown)


tarkah:
Should store models be managed from a separate repository where a wider group of individuals can have collaborator access / make quicker approvals to changes? It'd also be great if these models could be fetched / cached at runtime so that users don't need to constantly update to have fixes to broken links or new products.

jef:
> Should store models be managed from a separate repository where a wider group of individuals can have collaborator access / make quicker approvals to changes? It'd also be great if these store models could be fetched / cached / refreshed at runtime so that users don't need to constantly update to have fixes to broken links or new products.

Yes, something along these lines are ideal.

I've thought about spinning up an API that people can vote on/contribute with new links. I haven't spent any time doing these because it seems my free time has dwindled.

I do plan on this happening though!

jef (approved):

jef:
Thank you, @poulet42!

PR #2101, opened 2021-03-07
feat(pushover): ability to specify device

picchioni:
### Description

1) Adds the ability to specify which registered devices to send pushover notifications to.
2) Ordered the pushover variables in `dotenv-example` to approximate an order of importance (also matches documentation).
3) When adding the new variable to documentation, I slightly tweaked it as a potential example of how to document ones in the future. I included a link to the feature's API documentation, and a 3rd column to note whether specific variables are required for the feature to work. If you don't like it or would prefer different formatting/syntax let me know and I can revert or modify.

### Testing

Add the following to `.env` to send to specific devices
`PUSHOVER_DEVICE=device1,device2, etc...`

Leaving `PUSHOVER_DEVICE=` unpopulated will send to all registered devices, to match the current behavior.

### New dependencies

None


jef (approved):
Great! Looks good to me.

Thanks for the contribution 😁

Issue #1901, opened 2021-02-06
Asus store reports in stock cards but no stock is shown in browser.


LFAdams:
## Expected Behavior

The program should not report in-stock from the asus store when none are available.

## Current Behavior

the program constantly reports in-stock cards from the asus store, including pictures showing an add to cart button. When I click the link to go to the site there are none available. This happens every couple of minutes per card such that I get a constant stream of in stock notifications for asus, so i know im not just missing the stock by being slow.

## Steps to Reproduce

Install streetmerchant via dock container on unraid
include asus in STORES variable
Set notifications to use a gmail account
Receive endless in stock notifications for asus cards.
## Environment

- OS:
Unraid

```dotenv
```

bakamano:
also running into same issue

zkghost:
the pages show "add to cart" for a moment and then it reverts to "unavailable" or something, it's probably a cheap anti bot measure. Or rather makes the bot too spammy. i'd have to look at the code to see how easy this would be, but we need to essentially sleep for a few seconds and then check the state

PacketShepard:
Same issue here. I've disabled Asus stores for a few weeks now.

MuhsinunC:
Yup same issue here. @zkghost yes that is the issue. Any updates?

Fmstrat:
Checking on an update as well, running into this, too.

Fmstrat:
I think we need a delay option in the store `ts` that pulls through to `lookup.ts`. This way the anti-bot measure is thwarted. I thought at first this had something to do with the coming soon language, but that doesn't seem to be the case.

Fmstrat:
Anyone here willing to test this out, please let me know how it worked.

Fmstrat:
So this was tested and did not work. I'm wondering if somehow it is because the Asus config is out date. I'm seeing a number of cards who's model numbers have changed. I'm working through a PR on that and will see if it has any impact.

Fmstrat:
Well, that seems to be it. The above change worked for me (for 3080 searches). The 3060s are not up on their site at the moment, so ymmv there.

PR #2124, opened 2021-03-11
NEEDS TESTING: Added lookupDelay with ASUS configuration

Fmstrat:
### Description
The Asus store uses an anti-bot feature to show "add to cart" briefly before changing to an out of stock button. This fix circumvents that by delaying a configured amount of time per store (when specified). This is an untested PR, but can serve as a foundation for a fix if it does not function out of the box.

Fixes: https://github.com/jef/streetmerchant/issues/1901

### Testing
Run app, ensure calls to asus.com do not report false positives.

### DIFF
--- a/src/store/lookup.ts
+++ b/src/store/lookup.ts
@@ -408,6 +408,10 @@ async function checkIsCloudflare(store: Store, page: Page, link: Link) {
 }
 
 async function lookupCardInStock(store: Store, page: Page, link: Link) {
+  if (store.lookupDelay) {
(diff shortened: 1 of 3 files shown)


tseanard:
The code changes on their own in that commit generate 143 errors (due to not propagating the lookupDelay through everywhere VSCode thinks it should go.  

With that said, I tried hard coding a 3 second and a 10 second delay in the file with if (store.lookupDelay) { ... }, and even with a 10 second delay I am still getting false hits on the Asus store

PR #2131, opened 2021-03-13
refactor(store): replace euroFormat with regex

meyervp:
### Description

   remove useless parameter euroformat and replace it with autodetection by Regex
   fixes issue with european stores using dot notation for pricing

### Testing
   Tested on : 
        LDLC.com
        rueducommmerce.fr
        topachat.com
        amazon.fr
        newegg.com

### DIFF
--- a/src/store/includes-labels.ts
+++ b/src/store/includes-labels.ts
@@ -125,9 +125,8 @@ export async function getPrice(
   const priceString = await extractPageContents(page, selector);
 
   if (priceString) {
-    const priceSeparator = query.euroFormat ? /\./g : /,/g;
     const price = Number.parseFloat(
-      priceString.replace(priceSeparator, '').match(/\d+/g)!.join('.')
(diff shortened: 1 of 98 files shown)


jef (approved):
Great! This makes sense to me.

Thanks for the contribution!

tarkah:
This change seems to break when you have thousands separator AND decimal separator...

```
source from amazon id="priceblock_ourprice": $1,799.00

replace = $1,799.00 (doesn't actually do anything?)
match = ["1", "799", 00")
join = 1.799.00
parseFloat = 1.799
```

Also, the regex shouldn't be "escaped". It's actually doing nothing (matching on literal `\` and `.` or literal `\` and `,`.

tirrorex:
Anyone has a fix that they could recommend?
Because the current implementation (so after revert) is broken on a lot of European stores using the . instead of , to separate cents from euros.

PR #2145, opened 2021-03-15
refactor: move lowBandwidth parameter into store class to manage parameter by store

meyervp:
### Description

The parameter must be determined on an store basis and not be set by end-user (otherwise the bot consumes too much bandwith for no reason).

Set by default to "false" for all current stores.
Requires testing on all stores to determine if their parameter has to be set to "true".

The changes optimize crawling speed, bandwidth and cpu load.

### Testing

I tested and set on :
ldlc.com
topachat.com


jef:
I don't know if I like this change enough to pull it in. I do understand what you're suggesting and I think I'd rather make `LOW_BANDWIDTH=true` by default, rather than `false`.

bramevo:
> I don't know if I like this change enough to pull it in. I do understand what you're suggesting and I think I'd rather make `LOW_BANDWIDTH=true` by default, rather than `false`.

I can add a pull request for optimised LOW_BANDWIDTH setting. For example captcha you always want to get, so I've added an exception for that one. And I'm also blocking things like javascript/css/media/... Tested this on amazon and it works, but to know if it does not break any shop, it should be tested on all of them.

Default should indeed be true for LOW_BANDWIDTH.

Issue #2151, opened 2021-03-15
Regression from #2131


tarkah:
Related to #2131

This change seems to break when you have thousands separator AND decimal separator...

```
source from amazon id="priceblock_ourprice": $1,799.00

replace = $1,799.00 (doesn't actually do anything?)
match = ["1", "799", 00")
join = 1.799.00
parseFloat = 1.799
```

Also, the regex shouldn't be "escaped" here. It's actually doing nothing (matching on literal `\` and `.` or literal `\` and `,`. However, even if it was replacing correctly, the number would parse as `179900` which is also incorrect.

_Originally posted by @tarkah in https://github.com/jef/streetmerchant/issues/2131#issuecomment-799815951_

starcraft66:
I think I just ran into this issue. The latest nightly of streetmerchant is telling me that [this card](https://www.newegg.ca/evga-geforce-rtx-3080-10g-p5-3885-kr/p/N82E16814487520) is selling for $2.498 when it is in fact for sale for $2,498.99 on the website.

tarkah:
#2153 has a fix. However, I feel like the old way of defining the price in the model as "euro" and having much simpler logic that we _know_ will work, was the better solution. There is still a "hack" in #2153 that is sane, but will still potentially miss things.

jef:
Agreed @tarkah. I might revert if it continues to cause problems. I'll close this for now and reopen if we see this becoming a problem.

jef:
I ended up reverting everything, didn't want to hold this off much longer. We can revisit this when I have more time to test.

That being said, I have mocha and sinon as dev dependencies. Something like this in the future should really have testing around it.

Refs: 0ff8158511f4f90dc26407fffa22ab0a4ad3aa0b, 91c4f12308905f4e6d65341ba54ae814c8ad94d6

PR #2153, opened 2021-03-16
fix: parsing prices with thousands separator and European format

Ynng:
### Description

Fixed parsing prices with thousands separator and European.
The problem was introduced by #2131 and described here #2151
Example: Newegg Canada's price for RTX 3070 "1,600.00" was interpreted as 1.6 

### Testing

Tested manually and with a variety of stores around the world.

### DIFF
--- a/src/store/includes-labels.ts
+++ b/src/store/includes-labels.ts
@@ -116,16 +116,6 @@ export function includesLabels(
   );
 }
 
-function getPriceFromString(priceString: string, euroFormat: boolean): number {
-  return Number.parseFloat(
-    priceString
-      .replace(/\\/g, '')
-      .replace(euroFormat ? /\./g : /,/g, '')
-      .match(/\d+/g)!
-      .join('.')
-  );
-}
-
 export async function getPrice(
   page: Page,
   query: Pricing,
@@ -135,12 +125,16 @@ export async function getPrice(
   const priceString = await extractPageContents(page, selector);
 
   if (priceString) {
-    const euroFormat = priceString.indexOf('.') < priceString.indexOf(',');
-    let price = getPriceFromString(priceString, euroFormat);
-    //Sanity check, flip "euroFormat" if price lower than 50
-    if (price < 50) {
-      price = Math.max(price, getPriceFromString(priceString, !euroFormat));
-    }
+    const thousandsSeparator =
+      priceString.search(/\d+\.\d{3}|\d+,\d{2}$/) > -1 ? /\./g : /,/g;
+    const price = Number.parseFloat(
+      priceString
+        .replace(/\\/g, '')
+        .replace(thousandsSeparator, '')
+        .match(/\d+/g)!
+        .join('.')
+    );
+
     logger.debug('received price', price);
     return price;
   }
--- a/src/store/includes-labels.ts
+++ b/src/store/includes-labels.ts
@@ -116,6 +116,16 @@ export function includesLabels(
   );
 }
 
+function getPriceFromString(priceString: string, euroFormat: boolean): number {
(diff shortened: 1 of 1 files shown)


Ynng:
The new pull request (vs the old #2152) introduced a price sanity check.

It's near impossible to determine if a price string is in europian format (", and . flipped") or otherwise with just the string itself.
For example, `1,600` could be 1600 or 1.6 in europian format.

Now, if the price is lower than a hardcoded sanity check of `50`, "eurpian format" would toggle and the maximum price will be selected as the actual price.

In my testing, this seems to be working well.

meyervp:
my bad ! :/

another solution is use a recursive replace

src/store/includes-labels.ts
```diff
@@ -116,6 +116,16 @@ export function includesLabels(
...
 
+function getPriceFromString(priceString: string, euroFormat: boolean): number {
+  return Number.parseFloat(
+    priceString
+      .replace(/\\/g, '')
+      .replace(euroFormat ? /\./g : /,/g, '')
+      .match(/\d+/g)!
+      .join('.')
+  );
+}
```
jef, line 127:
```suggestion
function getPriceFromString(priceString: string): number {
  let price: number
  const euroFormat = priceString.indexOf('.') < priceString.indexOf(',');

  price = Number.parseFloat(
    priceString
      .replace(/\\/g, '')
      .replace(euroFormat ? /\./g : /,/g, '')
      .match(/\d+/g)!
      .join('.')
  );
  
  //Sanity check, flip "euroFormat" if price lower than 50
  if (price < 50) {
    price = Math.max(price, getPriceFromString(priceString, !euroFormat));
  }
  
  return price;
}
```

src/store/includes-labels.ts
```diff
@@ -125,10 +135,12 @@ export async function getPrice(
...
-    const price = Number.parseFloat(
-      priceString.replace(/\\.|\\,/g, '').match(/\d+/g)!.join('.') // eslint-disable-line
-    );
-
+    const euroFormat = priceString.indexOf('.') < priceString.indexOf(',');
+    let price = getPriceFromString(priceString, euroFormat);
+    //Sanity check, flip "euroFormat" if price lower than 50
+    if (price < 50) {
+      price = Math.max(price, getPriceFromString(priceString, !euroFormat));
+    }
```
jef, line 143:
```suggestion
    let price = getPriceFromString(priceString);
```

jef (dismissed):
Awesome! Thanks for this work.

I think it might be best to keep in contained into one function to reduce the cognitive load.

src/store/includes-labels.ts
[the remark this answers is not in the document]
Ynng: That sounds like a great idea, I'll try it out!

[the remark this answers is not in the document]
Ynng: Seems to be working great, thanks!

Ynng: The "Sanity check" is now removed in fc3c32f and so separate function is no longer needed.

Ynng:
The "Sanity check" is now removed in fc3c32f and so the separate function is no longer needed.
The function now checks for the thousands separator with this regex from @fongoid, which they explained in more detail [here](https://github.com/jef/streetmerchant/pull/2153#discussion_r595355043): 
`const thousandsSeparator = priceString.search(/\d+\.\d{3}|\d+,\d{2}$/) > -1 ? /\./g : /,/g;`

jef (approved):
Looks good to me!

Thank you.

nixeens:
Is this going to work if i have one store where price is showed like 1&nbsp;299,00 €

PR #2155, opened 2021-03-16
fix(asus): update urls

Fmstrat:
### Description
URLs were out of date for the ASUS US store. I left the 3060 URLs as is since they seem to have been temporarily removed from the store. This also has fixed https://github.com/jef/streetmerchant/issues/1901 for me when searching for 3080. I constantly got false positives, my guess is from one of the removed products.

### Testing
Ran searches using the new URLs. Manually validated each 3070/3080/3090 URL.

### DIFF
--- a/src/store/model/asus.ts
+++ b/src/store/model/asus.ts
@@ -66,10 +66,10 @@ export const Asus: Store = {
     },
     {
       brand: 'asus',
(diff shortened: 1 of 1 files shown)


jef (approved):
Great! Thank you.

PR #2182, opened 2021-03-18
refactor(smartthings): replace with node-fetch

jef:
### Description

Replace `@bridgerakol/samsung-smart-api` with `node-fetch` implementation.

This helps resolve vulnerability as our implementation isn't complicated enough to use the library and keep the vulnerability. The original author also doesn't have their code on GitHub, so reporting issues is not viable.


jef:
@pmilano1, are you able to verify this for me? I don't have anything Smart Things related and I'd like to get rid of this dependency.

PR #2239, opened 2021-03-27
chore(gamestop): update inventory

Reid-Kornman:
### Description

Gamestop started selling graphics cards, so I update the entire catalog, and fixed 2 broken xbox entries.

### Testing

verified all return proper statuses with npm run start


jef (approved):
Great! Thanks for taking care of this.

PR #2285, opened 2021-04-04
feat(notification): add support for Free mobile

aradenac:
### Description
Allow a user of Free mobile (french carrier) to receive SMS notifications.

The existing sms notification service could not be used, as the Free mobile notification service API does not
require a phone number. Hence PHONE_NUMBER is irrelevant in this context and two dotenv vars have been created.

### Testing
Tryed it with my own Free account, running 

> $ npm run test:notification


jef:
Interesting. So from what I understand, this mobile company doesn't require you to have a phone number, but a user can still receive SMS?

Strange... But OK!

jef (approved):
Great! Well thanks for the addition :)

aradenac:
> Interesting. So from what I understand, this mobile company doesn't require you to have a phone number, but a user can still receive SMS?
> 
> Strange... But OK!

Exactly, their web API takes a user ID and a on-purpose generated key only for the SMS notifications service... Like nobody else I guess. Free is often different than the others carriers (e.g. they where the first allowing users having a static public IP for home server at no cost). Provided their "home-automation" services come with no additional charges, they are often a good choice for hobbyists ^^

jef:
Aha, very interesting. Well, it'll sure help out you and others!

Thanks for the support.

PR #2290, opened 2021-04-05
feat(series): support for rx6700xt

aradenac:
### Description
As some users asked for this feature in the [discussion](https://github.com/jef/streetmerchant/discussions/2082), here is a proposal to support it.

### Testing
Tested with one example link for ldlc.fr, provided as an example so other can add their favorite stores later.


jef (approved):
Looks good to me! Thanks for the initial implementation!

PR #2311, opened 2021-04-07
feat/Colored empty stores

c4dx:
### Description

Applies to the `selected stores` information on startup.

Grayes out stores that have none of the selected brands/series/models as links. Additionally, if said even occurs, prints a line describing that some of the selected stores dont have what the user is looking for.

Prevents users (like me) from thinking the bot wouldn't work correctly because it isn't checking specific stores.

### Testing

Tested across series, models, brands for 3070 and 3080 on alternate and amd. Successfully grayed out not matching stores.

### New dependencies

None.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "twilio": "^3.59.0",
         "twitch": "^4.5.2",
         "twitch-auth": "^4.5.2",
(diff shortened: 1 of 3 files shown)


c4dx:
Should've checked the docs first.

PR #2321, opened 2021-04-08
chore(amd): add rx6800xt midnight black

davidramiro:
### Description

AMD released a new version of the reference RX6800 XT on their shop yesterday. I have added the links to all relevant `amd` stores.

### Testing

Created a new model for it in `store.ts` and made sure that specific link gets tested. Reports back stock correctly. For upstream, I left the model at `amd reference`, as it's just a black variant of the reference card.


jef (approved):
Great! Thank you.

Issue #2252, opened 2021-03-29
Drop happened last week and I got no notification


jonalange:
## Expected Behavior

Receive notification via discord

## Current Behavior

No notification received.

## Environment

- OS: Windows 10 Professional

Last week on 23.03 was a 3080 FE drop at notebooksbilliger.de (NBB) but I got no notification. Are there any logs to see what happened in the software? I can't find them.

Also, is there a way to check what page get checked when I select NBB? Maybe it's the wrong one.

hickneller:
You can see what page is being checked by setting HEADLESS to false.

PR #2352, opened 2021-04-13
fix(discord): properly wait for webhooks

crwgregory:
### Description
This commit adds promise resolution and objects to
be properly used within Promise.all function.
Before the `try {} catch {}` block would be exited
before the clients resolved their requests causing
uncaught errors from the notifications.
Also added object indexing to the notifyGroupSeries variable
to prevent typescript compile error.

Potentially Fixes:
- https://github.com/jef/streetmerchant/issues/2252

I failed to receive one of my alerts because of a uncaught error emanating from the discord notification webhook.

```
streetmerchant    | [6:52:25 PM] info :: 🚀🚨 [amazon] [{series} ({model})] ... :: IN STOCK 🚨🚀
streetmerchant    | https://www.amazon.com/gp/aws/cart/add.html?ASIN.1=...&Quantity.1=1
streetmerchant    | [6:52:25 PM] info :: ✔ discord message sent <--- Here the app is resolving out of the `try catch` block
streetmerchant    | /app/node_modules/discord.js/src/rest/RequestHandler.js:93 <--- Then fails to handle the 500 resp
streetmerchant    |         throw new HTTPError(error.message, error.constructor.name, error.status, request.method, request.path);
streetmerchant    |               ^
streetmerchant    |
streetmerchant    | HTTPError [FetchError]: request to https://discord.com/api/v7/webhooks/.../...?wait=true failed, reason: getaddrinfo EAI_AGAIN discord.com
streetmerchant    |     at RequestHandler.execute (/app/node_modules/discord.js/src/rest/RequestHandler.js:93:15)
streetmerchant    |     at runMicrotasks (<anonymous>)
streetmerchant    |     at processTicksAndRejections (node:internal/process/task_queues:94:5)
streetmerchant    |     at async RequestHandler.push (/app/node_modules/discord.js/src/rest/RequestHandler.js:39:14) {
streetmerchant    |   code: 500,
streetmerchant    |   method: 'post',
streetmerchant    |   path: '/webhooks/.../...?wait=true'
streetmerchant    | }
```

The `500` error I received might have been caused by the client being destroyed before the requests were resolved as I don't believe that `Promise.all` taking in an array of objects will know how to resolve the promises. These changes explicitly handle the promises and resolutions/catching of each webhook client.

### Testing
`npm run test:notification`
You can see that the logger is not called until the response is received by the client for each webhook.

### DIFF
--- a/src/notification/discord.ts
+++ b/src/notification/discord.ts
@@ -54,7 +54,6 @@ export function sendDiscordMessage(link: Link, store: Store) {
         const notifyKeys = Object.keys(notifyGroupSeries);
         const notifyIndex = notifyKeys.indexOf(link.series);
         if (notifyIndex !== -1) {
-          // notifyGroupSeries.
           notifyText = notifyText.concat(notifyKeys[notifyIndex]);
         }
 
--- a/src/notification/discord.ts
+++ b/src/notification/discord.ts
@@ -54,27 +54,36 @@ export function sendDiscordMessage(link: Link, store: Store) {
         const notifyKeys = Object.keys(notifyGroupSeries);
         const notifyIndex = notifyKeys.indexOf(link.series);
         if (notifyIndex !== -1) {
-          notifyText = notifyText.concat(Object.values(notifyGroupSeries)[notifyIndex]);
+          notifyText = notifyText.concat(
+            Object.values(notifyGroupSeries)[notifyIndex]
+          );
         }
 
-        let promises = [];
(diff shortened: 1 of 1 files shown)


jef (approved):
Ah, yes. This definitely is a lot better. I hadn't promisified the callback in `client.send`, so this would definitely cause this exception!

Thanks for looking into this and fixing it! :rocket:

PR #2371, opened 2021-04-15
fix(amazon-es): use euroFormat

AlexP11223:
It was parsing 1800 as 1.8 resulting in unwanted notifications. Now it seems to work correctly.


jef (approved):
Make sense, thanks!

PR #2375, opened 2021-04-16
fix(web): add favicon to prevent missing icon console error

Guy293:
fix: Add an icon of the project to the web ui to prevent this error:
`error :: Error in WebUI stream ENOENT: no such file or directory, open '\streetmerchant\web\favicon.ico`

### Alternative Approach

Another approach to this issue is specifying in the html, but adding it seems easier, cleaner and more sensible approach.


jef (approved):
Seems good to me! Odd that it was missing to begin with. Thanks for adding!

PR #2382, opened 2021-04-17
fix: partial url regexp

AlexP11223:
It was not working with relative URLs not starting with /
Now it determines url as partial if it does not start with http: or https:


AlexP11223:
So what's about this?

jef (approved):

jef:
Nice! Thank you :)

PR #2383, opened 2021-04-17
fix: Only set timeouts for active stores

mattrunyon:
### Description

Update the start function to read from the filtered list of stores instead of the entire list of stores. This reduces the number of timers that are created. With over 150 stores, most setups would have more than 90% be no-ops which have timers for no reason.

Also added types for the stores, series, models, and brands sets/maps since TS doesn't infer the type when it is created as an empty set/map. The type for the stores map is required due to the changes. The others aren't, but it is better than the inferred Set<any> type.

### Testing

Ran using debug mode with a set of stores in config. Prior to this change, you will get debug messages for every store in the project. After this change you should only see debug messages for the active stores.


jef:
This is great -- although it comes with caveats. I believe I also wrote this out like this at first, because of everything you said! But then @Doridian reminded me that this has side effects with `/web`. I forget all the reasons why exactly, but maybe they can explain better.

If there isn't, then we should definitely bring this in!

jef:
Related: https://github.com/jef/streetmerchant/pull/860, https://github.com/jef/streetmerchant/pull/680

Doridian:
So, yeah. These "wasteful" timeouts are needed as a "cheap" solution to be able to dynamically enable and disable stores using the WebUI.
To make the WebUI work for enabling/disabling, you would need to, on a change submitted through it, check the timers you already  have (keep them in a hashmap or similar), check the timers you need (new enabled stores), then cancel/start as needed.
Otherwise stores you enable would just never run until a restart of the whole app.

mattrunyon:
Gotcha. I can change this to just put the debug statements in the right place then because currently it will say it's checking every store if you enable debug log level, even if it's skipped.

Or change to keep a map of store name to timer for the web so stores can be enabled/disabled on the fly. That way the debug logs aren't flooded by disabled stores.

Doridian:
The map would be the cleanest solution, obviously. But if all you're concerned about is debug statements,  moving them is fine as well. Depends how much time you want to invest.

AlexP11223:
>  using the WebUI.

Does anyone use it? It seems quite limited. Also a bit confusing what will happen on the next restart: iirc it will just use dotenv again, so not clear what's the use case for such temporary update.

Doridian:
> > using the WebUI.
> 
> Does anyone use it? It seems quite limited. Also a bit confusing what will happen on the next restart: iirc it will just use dotenv again, so not clear what's the use case for such temporary update.

I used it quite a lot (which is why I PR'd it in). To constantly tweak my snatcher to take in/out of rotation stores when they CAPTCHA'd me, to let them cool down etc. Or to enable/disable cards once my friends found them.
My procedure was usually edit dotenv, then make the same changes in WebUI so I don't have to restart snatcher to apply dotenv.

mattrunyon:
So this actually should work with the web UI because of how Node modules behave like singletons (they're not exactly singletons, but in this case I think the stores object behaves as a singleton). I adjusted so that if a store is deselected, the next link lookup should immediately break out of the lookup loop and prevent the `setTimeout` from triggering again.

I made 1 log a debug and 1 an info (saying that the store is deselected)

Still need to add support for if a store is added via web UI. Right now should work with items added/removed, but only with stores removed. Stores added should be a bug in the current state of this PR

PR #2419, opened 2021-04-22
fix(alternate): use button selector instead of text

AlexP11223:
it often says "in stock" while also saying that it cannot be purchased instead of showing the cart button


jef (approved):
Much better. Thanks!

PR #2423, opened 2021-04-23
fix(nodemon): use npx with ts-node

mattrunyon:
### Description

Didn't find an issue reported, but the existing nodemon config required ts-node to be installed globally. Changed to run w/ npx so it will use the dev dependency version of ts-node installed

### Testing

`npm run start:dev` works. I believe if you run `npm rm -g ts-node` then the previous version will not work, but with this fix the dev start script will still work.


jef (approved):
Makes sense. Thanks for the update.

PR #2424, opened 2021-04-23
refactor(captcha): use seconds instead of ms

klords:
### Description

Nav link was missing for the new captcha reference page, and added some important usage info that was only in the PR last time around (how to respond successfully to the bot). 

Also, updated captcha timeout and poll interval values to be seconds instead of milliseconds. This was done for the sake of consistency, as I noticed other duration config values were specified in seconds as well.

### DIFF
--- a/docs/reference/captcha.md
+++ b/docs/reference/captcha.md
@@ -7,7 +7,14 @@ A mechanism has been implemented to allow users to interactively handle captcha 
 
 ## How to use
 
-To use this feature, simply configure the variables shown below and then start your streetmerchant instance.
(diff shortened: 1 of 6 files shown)


jef (approved):
Great -- I noticed the documentation was missing last night, so this is good.

Thanks!

AlexP11223:
> Also, updated captcha timeout and poll interval values to be seconds instead of milliseconds. This was done for the sake of consistency, as I noticed other duration config values were specified in seconds as well.

hm, isn't it the opposite?

e.g.

> Minimum sleep time between queries of the same product page. In **milliseconds**, default: 5000

klords:
The in stock wait time variable is defined in seconds. I was under the
impression others were in seconds as well, but don't recall what.

I would prefer every value was defined consistently, but I also don't want
to change the config usage too much either. Considering the poll interval
and timeout will probably never be sub-second intervals, leaving as seconds
is probably fine.

PR #2432, opened 2021-04-24
fix(microcenter): add outOfStock label

### LINKED ISSUE #2407: Microcenter in stock false positive
## Expected Behavior

When out of stock or 0 stock, should not be considered in stock and no alert triggered.

## Current Behavior

When the page says "0 NEW IN STOCK", the app considers this to be "in stock" and triggers an alert.

## Steps to Reproduce

Setup "microcenter" for a store, and look for 3080 series. It looks like Microcenter might sometimes show "0 NEW IN STOCK" or "OUT OF STOCK" interchangeably.

## Environment

- OS: Windows

## Logs

[9:45:20 AM] info :: 🚀🚨 [microcenter] [zotac (3080)] trinity :: IN STOCK 🚨🚀
https://www.microcenter.com/product/628607/zotac-geforce-rtx-3080-trinity-overclocked-triple-fan-10gb-gddr6x-pcie-40-graphics-card/?storeid=055

image

dlew:
The current inStock label only looks for the words "in stock"; however, recent changes to Micro Center's website results in there sometimes being "0 NEW IN STOCK".

Refs #2407

### Description

Fixes #2407

### Testing

Tested by setting `STORES=microcenter` then running.

Before the fix, it would show IN STOCK for everything, even when it was out of stock. With this branch, it shows OUT OF STOCK for everything (since that's the current state of the world as of testing).

### DIFF
--- a/src/store/model/microcenter.ts
+++ b/src/store/model/microcenter.ts
@@ -347,6 +347,10 @@ export const MicroCenter: Store = {
       container: '#pnlInventory',
       text: ['in stock'],
     },
+    outOfStock: {
+      container: '.inventoryCnt',
+      text: ['0'],
+    },
     maxPrice: {
       container: 'span[id="pricing"]',
       euroFormat: false,


jef (approved):
Looks good to me! Thanks!

PR #2437, opened 2021-04-25
feat: add waitUntil and waitForSelctor to linksBuilder

AlexP11223:
I was trying to add a linksBuilder for some store, but the store loads the search results via ajax, so they were not in the response passed to linksBuilder.

Added two new options for linksBuilder:

- `waitUntil` - the same as in Store. allows to specify one of the `PuppeteerLifeCycleEvent` values.
- `waitForSelctor` - wait for the specified element to appear, using puppeteer's `page.waitForSelctor()`.


jef (approved):
Cool! I like the idea. Thanks for the addition 😁

PR #2460, opened 2021-04-28
fix(captcha): use Slack captcha handler token when provided

### LINKED ISSUE #2428: Slack ignores CAPTCHA_HANDLER_TOKEN
Looks like Slack always uses token from `SLACK_TOKEN` ignoring `CAPTCHA_HANDLER_TOKEN`.

https://github.com/jef/streetmerchant/blob/d652b4936435377d36268141a2b63c8798f7f95b/src/messaging/slack.ts#L6-L7

I guess it should either try `CAPTCHA_HANDLER_TOKEN` first when calling `getSlackCaptchaInputAsync`, and then maybe fallback to `SLACK_TOKEN` if empty, or the documentation needs to be updated.

@klords

klords:
### Description

Fixes #2428 

Uses the captcha handler token provided and tries to fall back to `SLACK_TOKEN` if missing. This approach allows a user to have a distinct bot for captcha vs notifications, if they so choose. (I've opted to not pre-emptively generalize this for other messaging services.)

### Testing

`npm run test:captcha` with:
- [X] only `CAPTCHA_HANDLER_TOKEN` defined
- [X] only `SLACK_TOKEN` defined
- [X] both defined
- [X] neither defined


jef (approved):

jef:
Great! Thanks @klords for this 😃

Issue #2369, opened 2021-04-15
Crashes on notification


FrankFioretti:
## Expected Behavior

Notification should be sent and script should continue to run

## Current Behavior

Notification is sent and script crashes/exits UNLESS running as local administrator on Windows 10. Is this expected?

## Environment

- OS:

<details>
<summary>dotenv</summary>

```dotenv
# Place your dotenv within these backticks. Be sure to remove any secrets/passwords.
```

</details>

FrankFioretti:
Note: still crashes intermittently when running as administrator

MieszkoT84:
I think it is the same behaviour as in this topic: https://github.com/jef/streetmerchant/issues/2232
As of now I don't know how to completely fix it but there is workaround which has to do for the time being. Please see the topic above for details.

Austin-Wells:
Having the same issue, crashes immediately after opening a chromium instance. Powersehll/Git still running in process manager.

Eribaeri:
i got the same problem

jef:
Closed with https://github.com/jef/streetmerchant/pull/2461

klords:
Note to anyone following this issue: the merged fix will only take effect when you (re)install node modules using `npm i`

MieszkoT84:
> Note to anyone following this issue: the merged fix will only take effect when you (re)install node modules using `npm i`

I would also add that 'git clone https://github.com/jef/streetmerchant.git' command is also necessary because 'npm i' didn't really work for me but after cloning the entire project again all is dandy now. Thanks a lot!

PR #2461, opened 2021-04-28
fix(lookup): prevent terminal from being hidden

### LINKED ISSUE #2232: Program closes when item is found
## Expected Behavior

Is this intentional - I would expect the program to keep running to find other items. 

## Current Behavior
[dotenv2.txt](https://github.com/jef/streetmerchant/files/6209369/dotenv2.txt)

Command prompt shuts down once item is found and notification sent.

## Environment

- OS:

Windows 10

klords:
### Description

Fixes #2232, #2369

Known issue starting in `open=^8.0.3`, details [here](https://github.com/sindresorhus/open/pull/238). Pinning to 8.0.2 for now will resolve this while the linked PR is approved.

### Testing

- open `cmd`, navigate to sm directory
- run `npm i` with updated `package.json`
- ensure `OPEN_BROWSER` is not `false`
- run `npm run test:notification`, verify browser window opens while `cmd` window also remains open
- (can also run against test item and verify `cmd` continues processing after opening browser when using `npm run start`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "node-notifier": "^9.0.1",
     "node-pagerduty": "^1.3.6",
     "nodemailer": "^6.5.0",
-    "open": "^8.0.6",
+    "open": "8.0.2",
     "play-sound": "^1.1.3",
     "puppeteer": "^9.0.0",
     "puppeteer-extra-plugin-adblocker": "^2.11.11",


deleto:
Works for me (Windows 10).

jef (approved):
Thanks for looking into this. I don't have a Windows machine always readily available, so I appreciate it!

PR #2462, opened 2021-04-28
feat(store): add amd-at

SoMuchForSubtlety:
### Description
Add support for AMD AT. Copy-pasted from AMD BE with the country path changed from `/be` to `/at`


jef (dismissed):
Thanks!

jef (approved):

PR #2463, opened 2021-04-28
feat(store): add cyberport-at

SoMuchForSubtlety:
### Description

Add support for [cyberport.at](https://www.cyberport.at/). The product IDs are the same as [cyberport.de](https://www.cyberport.de/), so I just had to replace `.de` with `.at`.


jef (approved):
Looks good to me! Thank you for the contribution! 💯

Issue #2452, opened 2021-04-27
Amazon-gb captcha not working


POPLOPUK:
## Expected behaviour

Once the street merchant detects a site with a captcha a DM is sent to the user specified and enter the correct captcha.

## Current Behavior

No message sent to user

## Steps to Reproduce

Add amazon-gb to shop and enable captcha

## Environment

- OS: Ubuntu-20.04

<summary>dotenv</summary>

```dotenv
# ** All configuration variables are optional **
# Copy this file and name it as `dotenv` (without backticks)
# Read https://jef.codes/streetmerchant/getting-started/#customization for help on customizing this file
########################################################################################################

ASCII_BANNER=
ASCII_COLOR=
APNS_AUTHKEY=
APNS_BUNDLEID=
APNS_DEVICETOKEN=
APNS_KEYID=
APNS_PRODUCTION=
APNS_TEAMID=
AUTO_ADD_TO_CART=
BROWSER_TRUSTED=
CAPTCHA_HANDLER_POLL_INTERVAL=
CAPTCHA_HANDLER_RESPONSE_TIMEOUT=
CAPTCHA_HANDLER_SERVICE=discord
CAPTCHA_HANDLER_TOKEN=[removed]
CAPTCHA_HANDLER_USER_ID=[removed]
DESKTOP_NOTIFICATIONS=
DISCORD_NOTIFY_GROUP=[removed]
DISCORD_NOTIFY_GROUP_3060=
DISCORD_NOTIFY_GROUP_3060TI=
DISCORD_NOTIFY_GROUP_3070=
DISCORD_NOTIFY_GROUP_3080=
DISCORD_NOTIFY_GROUP_3090=
DISCORD_NOTIFY_GROUP_CORSAIR_SF=
DISCORD_NOTIFY_GROUP_RX6800=
DISCORD_NOTIFY_GROUP_RX6800XT=
DISCORD_NOTIFY_GROUP_RX6900XT=
DISCORD_NOTIFY_GROUP_RYZEN5600=
DISCORD_NOTIFY_GROUP_RYZEN5800=
DISCORD_NOTIFY_GROUP_RYZEN5900=
DISCORD_NOTIFY_GROUP_RYZEN5950=
DISCORD_NOTIFY_GROUP_SONYPS5C=
DISCORD_NOTIFY_GROUP_SONYPS5DE=
DISCORD_NOTIFY_GROUP_XBOXSX=
[...]

PR #2468, opened 2021-04-29
feat(captcha): add support for configurable capture type

klords:
#### Update: The logic is now configurable and defaults to the current implementation `link` but is settable per-store or with a global override in dotenv. This sidesteps the overall impact of the below discussion points (but not its relevance to the available options).

Changes the logic of the captcha handler to take a screenshot of the target element rather extracting an image url. This approach has some advantages and drawbacks worth consideration:

| Pros | Cons |
|---|---|
| Prevents flagging from excessive or unusual downloads/accesses of captcha images | Counts against messaging service file upload quotas |
| "Future-proof" for when we want to take on harder captchas | We would have to use a strategy pattern approach anyways based on store config or other detection method |
| Resistant to using non-image-based captchas | Is this captcha tech in use? Would it be easier to solve anyway? |

I also added the captcha handler config for all Amazon sites after testing them in bulk. It appears they all have the same structure (as one might expect). Guess that'll fix issues like #2452.

### Testing

`npm run test:captcha` with various service configurations. Captcha test is updated to actually open a live page and grab image.


AlexP11223:
I would make an option for the captcha capture method. Like this:

```
async function saveCaptchaAsync(method: string, page: Page, store: Store): CaptchaPayload {
  switch (method) {
     case 'screenshot':
       return saveCaptchaAsync(page, store);
    case 'url' ... 
  }
}
```

Probably should be configurable per store, and maybe allow to override via env.

And maybe should rename `DMPayload` to `CaptchaPayload`, for adding captcha recognition services and libs (and I would extract all handlers from `messaging` to e.g. `captcha/handler`). For some captcha (such as recaptcha iirc) you may need to just extract some ID from HTML.

For saving files maybe easier to use tmp dir. https://github.com/benjamingr/tmp-promise

klords:
That's a good thought on making the capture type configurable, will
consider adding that.

Any changes in anticipation of adding another processing service (e.g. paid
captcha service) is premature. That refactoring can and should happen in
the PR that adds that support. I have no intention of adding it.

There are no captcha handlers in the messaging code. The captcha references
there can simply be renamed (aside from captcha.ts, which is meant as an
entry point to specific messengers for captcha messages), as they're
generic convenience methods for sending a DM and awaiting response. To that
end, also renaming from DM to Captcha payload is mostly not a good idea,
since those DM payloads are for... DMs.

AlexP11223:
>  There are no captcha handlers in the messaging code.

I mean just extracting this

https://github.com/jef/streetmerchant/blob/39fac90ba4feadb7e5c681db60e60a83749f5e62/src/messaging/discord.ts#L168

it seems like it can be outside of this module since it's only using the more generic exported stuff. And then all possible captcha handlers will be nicely in one place making it more clear.



> I have no intention of adding it.

I can do it :)

Also I saw [this lib](https://pypi.org/project/amazoncaptcha/) for Amazon captchas used in fairgame, but not sure if there are JS alternatives.

Though handling Amazon captchas actually seems not very important, because there is no chance to buy a GPU manually, there are too many buying bots, and even without bots thousands of people monitor Amazon.

jef:
Sorry @klords, I haven't gotten around to reviewing this.

Is this PR in it's final state?

jef:
Looking at this today @klords. Thanks for your patience and keeping this updated.

src/index.ts
```diff
@@ -28,48 +28,7 @@ async function restartMain() {
...
-    args,
-    defaultViewport: {
-      height: config.page.height,
-      width: config.page.width,
-    },
-    headless: config.browser.isHeadless,
-  });
-
-  config.browser.userAgent = await browser.userAgent();
+  browser = await launchBrowser();
```
jef, line 31: 👍

src/util.ts
```diff
@@ -104,3 +105,11 @@ export async function getRandomUserAgent(): Promise<string> {
   logger.debug('user agent', {userAgent});
   return userAgent;
 }
+
+export function deleteFile(path: string) {
```
jef, line 109: **suggestion(non-blocking):** As previously mentioned, it could be easily to put into a temp directory. But this is fine for now as I don't want to block this.

jef (approved):
Overall, looks great. I really appreciate the hard work you've put into this topic. This has been a great addition.

jef:
Change to `feat` to get captured by release-please.

PR #2482, opened 2021-05-03
fix: remove spaces from price

AlexP11223:
Some stores use space as thousands separator (`1 234.56`) and they were incorrectly parsed as `1.234`  passing the max price check when not supposed to.


jef (approved):
Make sense to me! Thank you.

PR #2483, opened 2021-05-03
feat(store): add multicom.no and more products to NO stores

davols:
### Description
- Add multicom.no as a store
- Add cartUrl for elkjop.no
- Add new links to existing norwegian stores

### Testing
Tested with products in-stock.


jef (approved):
Looks great! Thank you.

PR #2484, opened 2021-05-04
feat: add INCOGNITO to dotenv-example and volta support

Aragorn450:
### Description

- Added `INCOGNITO=` to dotenv-example since it was missing.
- Added support for [volta](https://volta.sh/) to package.json.
  - With this addition, the volta package manager will automatically switch to the correct version of node when trying to run node from this project. This eliminates the need to remember to change the version of node with nvm or other package managers first.

### Testing

Changed the default version of node on my local machine volta uses to 14.16.1 and then browsed to the streetmerchant directory. Checked the version of node with `node --version` and it showed `16.0.0` as expected.

### New dependencies

None


jef:
I currently use [fnm](https://github.com/Schniz/fnm), but plan on using [asdf](https://github.com/asdf-vm/asdf). Volta seems pretty cool for those who just want a node manager, but I'm trying to understand why it needs to be in the package. Does it not read the `.node-version`? I would imagine they would want to work with more popular systems (like [nvm](https://github.com/nvm-sh/nvm)).

jef (approved):

PR #2505, opened 2021-05-06
chore(elkjop): update selectors

davols:
### Description
Adds price selectors to elkjop

### Testing
Tested with 2 non-gpu items with higher than target price and lower than target price.


jef (approved):

jef:
Thank you!

PR #2515, opened 2021-05-10
chore(otto): add labels for xbox series x & s

Onind:
### Description

fixes false positives for the series x (screenshot shows the series s page after redirect)

### Testing

tested with docker local and the alert for series s still works. could not test series x because the page is currently hidden.


jef (approved):
Looks good to me! Thanks @Onind 😄

PR #2547, opened 2021-05-17
chore(netonnet-no): update selectors

davols:
### Description
Netonnet-no has updated their site, all the buttons now have ID which includes the SKU for the current item. 
I also added price label 

### Testing
Tested with out of stock GPUs and in stock items (including w/ price in dotenv)


jef (approved):

jef:
Thanks!

PR #2598, opened 2021-06-03
feat: add 3080ti

govizlora:
### Description
Add 3080, 3080ti and 3090 to bestbuy ca. Since 3080ti is new, corresponding types and documentation are also updated

### Testing
N/A

### New dependencies
N/A


jef (approved):
Great! Thank you!

PR #2694, opened 2021-07-29
fix(bestbuy-ca): avoid false positive

### LINKED ISSUE #2641: Bestbuy-ca False Positives
## Expected Behavior

Out of stock: Prompt says out of stock.
In stock: Prompt says its in stock, opens the webpage & adds to cart.

## Current Behavior

Out of stock: Prompt says its in stock & opens the webpage.
None of the gpus I'm looking for have been in stock since issue first occurred, so I don't know in stock behavior.

## Steps to Reproduce

Running v3.5.0

Happens ~50% of the checks on any gpu it searches for.

Ex. card:
https://www.bestbuy.ca/en-ca/product/msi-nvidia-geforce-rtx-3070-ventus-3x-oc-8gb-gddr6-video-card/15038016

Capture

I don't know much about HTML so this may be wrong, but I think they changed the add to cart button and the bestbuy-ca.ts is looking for a container that doesn't exist anymore?

## Environment

- OS:
Windows 10

<details>
<summary>dotenv</summary>

```dotenv
I reset to default config and changed only a few items to narrow down search, rest are default values.

LOW_BANDWIDTH=true
PAGE_SLEEP_MIN=100000
SHOW_ONLY_BRANDS=msi, EVGA, nvidia
SHOW_ONLY_SERIES=3060ti, 3070
STORES=bestbuy-ca

```

</details>

richie256:
## Description

Fixes #2641

Checking for inStock instead of the outOfStock element will make any error related to the page loading to be considered as "OutOfStock" instead of a false Positive.

## Testing

- I've run the check for all the elements of BestBuy-ca for 30 minutes and there were no false positive.
- I've run the Positive check using test:series item and it appeared as In-Stock.

### DIFF
--- a/src/store/model/bestbuy-ca.ts
+++ b/src/store/model/bestbuy-ca.ts
@@ -7,8 +7,8 @@ export const BestBuyCa: Store = {
       container: 'div[class*="pricingContainer"]',
       euroFormat: false,
     },
-    outOfStock: {
-      container: '.addToCartButton:disabled',
+    inStock: {
+      container: '.addToCartButton:not(:disabled)',
       text: ['add to cart'],
     },
   },


jef (approved):
Looks great!

Thanks :)

PR #2726, opened 2021-08-13
fix(xbox): false positive

VioSynthax:
### Description
Fixes false positives when using the Xbox store.

### Testing
Ran with fix for a week or so, checks for stock successfully after updating.


jef (approved):
Awesome. Thank you for doing this @VioSynthax.

Looks great!

PR #2734, opened 2021-08-19
chore(store): EBGames renamed to Gamestop (in Canada)

richie256:
### Description

EBGames has been renamed to Gamestop in Canada. I've removed `ebgames` and created `gamestop-ca` store. Also updated the store filter from the Docs.

### Testing

- Added `gamestop-ca` store in `dotenv` and tested with `sonyps5c`, `sonyps5de`.

- Tested positive result using `test:series`.


jef (approved):
Thank you!

PR #2850, opened 2021-11-15
chore(mediamarkt, saturn): add new outOfStock message

muffe:
### Description

Added new outOfStock message to prevent false alarm

### Testing

Ran the changes locally, the false alarm disappeared

PR #2853, opened 2021-11-16
chore(mediamarkt, mediamarkt-at, saturn) Additional Out-Of-Stock messages for MediaMarkt (AT/DE) and Saturn (DE)

prosellen:
chore: Adds new MediaMarkt (DE/AT) and Saturn (DE) have added out-of-stock messages

### Description
Like PR #2850, but with an added file for Saturn.de - all three sites are more or less different storefronts for the same store.

### Testing

Runs on my installation without false-positives.


jef:
Is this not relevant?

PR #2854, opened 2021-11-16
feat(store): add zotac queue detector, evga scraping

Mr-Tech-13:
### Zotac Queue Detector+Evga product page scraping+Newegg captcha handler

Updated zotacstore so that it detects a queue, Evga product pages can now be scanned, and newegg has a captcha handler
Screen Shot 2021-11-16 at 6 41 24 PM
Screenshot (9)


jef (changes requested):
Awesome! This is very much needed and will help a lot of folks.

If you could fix the linting and remove the duplicate label part, I think this can go in!

```
  labels : {
    inStock: {
        container: '#EVGAContent > div > div.main-wrapper > div.main-product',
        text: ['08G-P5-3665-KR'],
             },
            },
```

Mr-Tech-13:
Ok, done

src/store/model/evga.ts
```diff
@@ -15,108 +15,246 @@ export const Evga: Store = {
...
+  labels : {
+    inStock: {
+         container: '#EVGAContent > div > div.main-wrapper > div.main-product',
+         text: ['100-BR-0450-K1'],
+             },
+            },
+         model: 'test:model',
+         series: 'test:series',
+          url: 'https://www.evga.com/products/feature.aspx',
+ },**/
```
jef, line 29: Do you mind removing this if it's not in use?

src/store/model/zotac.ts
```diff
@@ -18,72 +18,16 @@ export const Zotac: Store = {
...
-    },
+        'https://zotacstore.queue-it.net/?c=zotacstore&e=zotacprod510440',
+    },
+   {
+    brand: 'zotac',
+    model: 'zotac',
+    series: 'zotac',
+    url:
+      'https://zotacstore.com',
+   },
```
jef, line 30: Can you explain how this would check if we're in a queue or if a queue is available? It seems that `You are now in line` doesn't make much sense to me.

jef (commented):
Just a couple questions.

Thanks!

src/store/model/evga.ts
Mr-Tech-13: Sure

src/store/model/zotac.ts
Mr-Tech-13: When there is a Zotac queue zotacstore.com redirects to the queue page and says "You are now in line"

Mr-Tech-13:
All done

jef:
I like the EVGA changes, but it seems that Zotac has products online. Am I missing information about the queue?

Mr-Tech-13:
The way that the Zotac Queue works is that when There is a queue then it will redirect you to a specified URL, which is why the link for the store is still the same, but the selector is different.

PR #2855, opened 2021-11-17
feat(redis): introduce pub/sub

gmalette:
#### Fix: Redis

The existing redis logging message did not behave properly. Redis commands are asynchronous. As such, the returned value of `client.set` was whether the command was buffered or not, and wasn't indicative of the success or failure of the command. This PR addresses that issue.

Additionally, the dotenv file was updated to show how to configure redis

#### Feat: Redis

Redis contains a pub/sub mechanism. I'm not sure how the key set was meant to be used, but it seems like the only way to know if an item becomes available is to poll while performing `keys *`, which is less than ideal. This PR starts publishing messages on the `"streetmerchant"` topic, to which any client can subscribe.

### Testing

I manually subscribed to the `streetmerchant` topic, ran streetmerchant on a test product, and confirmed that the message was published.


jef (approved):
Thank you!

PR #2914, opened 2022-01-13
fix: amazon prices

wallopthecat:
### Description
Amazon US was not pulling prices.

### Testing

Manually tested items are being filtered on Amazon by price


goldsziggy:
+1 and watching PR - disabling Amazon in the mean time as it creates false positives.

jef (approved):
Thank you :)

PR #2938, opened 2022-01-23
chore(store): update norwegian links

paalr:
### Description
Fixed links, added maxprice, added products, removed old products, and updated selectors on Norwegian stores


jef (approved):
Great! Thank you :)

PR #2944, opened 2022-01-27
feat: store: added 3050 in topachat.ts and materiel.ts

OscarleBG:
Hi, as 3050 went out today, I added 3 links to buy them in France.


OscarleBG:
I didn't try to run it at the time, and realized later it failed npm start because the series and models were not defined, I'll try to figure that out and PR later if I get it running.

PR #2945, opened 2022-01-28
feat(store): add 3050 in topachat and materiel

OscarleBG:
Hi, as 3050 went out today, I added 3 links to buy them in France.


jef (approved):
Looks good! Thanks :)

PR #3103, opened 2022-12-08
fix: rectify 4090 URLs in newegg.ts

bobbydp:
Removed redundant entries, and rectified URLs to point to their labeled resource. There were some inconsistencies between the Cart URL and the item number that the product belonged to.

### Description
There are URLs that were inconsistent with the current state of Newegg, or internally pointed at mismatched Item Numbers. I went through and revalidated each one.

### Testing
I tested each URL in browser and just ensured that the item numbers line up with their respective labels.

### DIFF
--- a/src/store/model/newegg.ts
+++ b/src/store/model/newegg.ts
@@ -519,8 +519,8 @@ export const Newegg: Store = {
     {
       brand: 'gigabyte',
       cartUrl:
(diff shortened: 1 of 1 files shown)


bobbydp:
Additionally, for the code hunks that are entirely removed, see above or below the removed hunk to reveal its respective redundant entry.

bobbydp:
I decided to cherry pick some of the parts of Mr-Tech-13's 4080 and 4090 additions to clean up that work (at least for the US stores) and merge them into this pull request. Standby for a different pull request...

PR #3198, opened 2025-01-29
feat: Add missing documention + env variables, add country attribute to stores, fix linting errors, deprecate Twitch notification

alanbixby:
### Description

While setting up my `streetmerchant` client ahead of tomorrow's 5090 launch, I noticed the documentation had fallen out of sync over the years and linting rules were not being enforced. As part of these changes, I ran an extract to update the list of models per series, add notify groups to missing products like the A3 mATX Lian Li case, and fixed some type errors + linting errors.

All of the models + series were code gen'd based on the exports in `src/store/model` and cross referenced with the new models in #3191 . As a future enhancement, this probably should be procedural and moved into the CI/CD pipeline - but that's out of scope for this PR.

Regarding linting rules, I have also disabled the linting rule `@typescript-eslint/no-explicit-any` - there library currently has ~40 linter warnings for this rule - all of which I'd consider acceptable uses of `any`.

--

> ⚠ Deprecation!

While testing with the `main` branch, I found the Twitch notification client was firing (despite blank environment variables!) on any product match and threw a runtime error on any product hit. 

```txt
var _this = _super.call(this) || this;
   TypeError: Class constructor EventEmitter cannot be invoked without 'new'
    at ChatClient.IrcClient [as constructor] (/streetmerchant/node_modules/.pnpm/ircv3@0.26.14/node_modules/ircv3/lib/IrcClient.js:24:28)
    at new ChatClient (/streetmerchant/node_modules/.pnpm/twitch-chat-client@4.6.7_twitch@4.6.7/node_modules/twitch-chat-client/lib/ChatClient.js:44:28)
    at Object.<anonymous> (/streetmerchant/build/src/messaging/twitch.js:23:20)
```

This is an error from an upstream dependency via `twitch-chat-client` - these packages are deprecated in favor of `@twurple/chat` and `@twurple/auth` - but given the timeline and my lack of experience with these libraries I opted to disable the functionality and add a warning in `config.ts` for users who provide a Twitch credentials. Given the lack of issues + the recently merged PRs, it is entirely possible this is an issue isolated to my environment, if that's the case let me know and I can revert that portion of the PR. 

--

I have also laid the groundwork for a `SHOW_ONLY_COUNTRY` filter. Implementation is not complete, but attributes have been added to all of the store objects, and a `getAllCountries()` function was written - I just haven't wrapped my head around how the web UI pushes updates or a clean place to integrate the filter in this PR.

--

[...]

PR #3199, opened 2025-01-30
feat: add MSI US storefront

alanbixby:
> ⚠depends on updates in #3198

### Description

This change adds all of the current GPUs on the https://us-store.msi.com/ storefront - consequently, `4060`, `4070`, `4070super`, and `4070ti` are newly added series types with their corresponding filter parameters as well

### Testing

image

image

image

### Other Changes

`config.ts` has been modified such so if `SHOW_ONLY_SERIES=` is blank, all series with defined in the notify config except `test:series` and `captcha-deterrent` are activated automatically - which should hopefully make this one less items to maintain when adding new series


src/store/model/msi.ts
```diff
@@ -0,0 +1,253 @@
+import {Store} from './store';
+
+export const MSI: Store = {
+  currency: '$',
+  labels: {
+    inStock: {
+      container: '#button-cart',
```
alanbixby, line 7: And for those of you camping out for the pre-order launch ([since MSI isnt shipping on launch day](https://videocardz.com/newz/msi-delays-geforce-rtx-5090-official-us-store-availability-to-february-6))

```js
inStock: {
    // Unknown what the pre-order text looks like so lets see what sticks!
    container:
      '#product_qty button:not([onclick^="compare.add"]):not([onclick^="notify.showNotify"])',
    text: [''],
  }
 ```
 should capture anything that isnt explicitly the existing compare or notify me buttons - this is too aggro for the production release though IMO

File: src/config.ts [hunk 2]
```diff
@@ -499,46 +507,12 @@ const store = {
...
-    'sonyps5de',
-    'udm-pro',
-    'udm-us',
-    'udr-us',
-    'unvr',
-    'unvr-pro',
-    'xboxss',
-    'xboxsx',
-  ]),
+  showOnlySeries: envOrArray(
```
jef, line 510: Much, much cleaner and clear 🥂 Thank you!

jef (approved):
*chef's kiss*

Thank you for the additional store and the awesome work that lead up to it!

xgalaxy:
This storefront doesn't seem to be working. I never see it show up in the logs while the other stores do.

PR #3207, opened 2025-01-31
fix: rickroll selector is broken for captcha test

joel-krueger:
### Description

Fixes the `npm run test:captcha` test.
The wikipedia has apparently changed the image class, such that img.thumbimage no longer selects anything on the page.

### Testing

If you run the test it throws a timeout exception as it is today.

My testing consisted of inspecting the rick roll element, clicking copy selector, and pasting it to replace img.thumbimage, and then running the test.

I'm not a node expert, or a selector expert, so feel free to decline this and change it to something more efficient / less likely to break as this is a very specific selector replacement.

### New dependencies
N/A


test/util.ts
```diff
@@ -39,7 +39,7 @@ export function getTestStore(): Store {
         text: ['Rickrolling'],
       },
       captchaHandler: {
-        challenge: 'img.thumbimage',
+        challenge: '#mw-content-text > div.mw-content-ltr.mw-parser-output > figure:nth-child(3) > a > img',
```
jef, line 42: This type of selector can be inconsistent, but thank you for updating and bringing it to my attention.

Try to aim for a specific selector (like the previous one) to avoid unintended interactions with other elements on the page.

```suggestion
        challenge: '.mw-file-element',
```

joel-krueger:
#3213 was opened for this issue

jef (approved):

PR #3210, opened 2025-01-31
Add 50xx NVIDIA Links to global stores

Mr-Tech-13:
Added links to NVIDIA-gb, NVIDIA-FR, and NVIDIA-DE


GewoonJaap:
Don't forget to run prettier ;)

EbonyEquivocal:
I noticed you've changed the URL from https://shop.nvidia.com to https://marketplace.nvidia.com; how come? Does this matter? 

Also, running your code, I get a lot of `[nvidia-gb] nvidia 5080 founders edition - Navigation timeout of 30000 ms exceeded` errors compared to before.

Mr-Tech-13:
It matters because store.nvidia.com no longer exists. Yes the timeout is normal. IDK Why it does it but the program is still 100% functional. It has something to do with the way Nvidia renders the site but if you run headless=false, you will see that the site is fully loaded the program just gets confused and hits the timeout.

EbonyEquivocal:
> It matters because store.nvidia.com no longer exists. Yes the timeout is normal. IDK Why it does it but the program is still 100% functional. It has something to do with the way Nvidia renders the site but if you run headless=false, you will see that the site is fully loaded the program just gets confused and hits the timeout.

Fair enough. Hopefully we can find a solution to the timeouts / errors.

src/store/model/nvidia-gb.ts
```diff
@@ -4,12 +4,12 @@ export const NvidiaGB: Store = {
   currency: '£',
   labels: {
     inStock: {
-      container: '.buy',
-      text: ['add to cart', 'buy now'],
+      container: 'span.buy-link-atc',
```
EbonyEquivocal, line 7: Are you sure that the container ID for a founders being in stock is that same as it being in stock for a third party card? I'm raising this because the button looks different between founders and third party cards. When I checked the element for an out-of-stock FE card and an in stock third party card, they're different as well.

RTX 5080 FE in stock
image

Third party card in stock
image

Mr-Tech-13: Feel free to fix it. I only designed it with fe cards in mind because as far as I could tell nvidia doesn't directly sell third party cards as far as I know. It's more efficient to track the third party

EbonyEquivocal: I'm the same, I only want an FE card. The container ID you have works for third party cards but isn't the same for out of stock founders card from what I can see. It's annoying because the the founders card isn't in stock, we can't check what the ID is for the button.

P-Adamiec:
Hello, I'm returning for my links too. Soo, as said, the official link for FE in my case is:
https://marketplace.nvidia.com/pl-pl/consumer/graphics-cards/nvidia-geforce-rtx-5090/ 
It is just as predicted. By replacing it to de-de I was able to investiage that the selectors have changed (If '.buy' means a class selector where there is this word in it - trying to learn  some code. However, I'm not confident to do it alone. So please add the Polish store too, as I'm awaiting for 5090 FE release in Poland as we got 0 on paper launch...
image

PR #3221, opened 2025-02-16
fix: update selector for newegg product price

mven:
### Description

Currently, the MAX_PRICE_* feature doesn't work with Newegg, because it's using the wrong selector. This change updates the selector so that the product price is correctly selected. I've also changed the `outOfStock` selector to `.product-buy` because `.product-inventory` does not exist on the product page.

### Testing

- I tested the Out of Stock functionality and works as it should.
- To test the max price feature, I manually changed a product link in the Newegg store that was in stock and set a MAX_PRICE. I've verified the notification is ignored if the product is over the max set.

### DIFF
--- a/src/store/model/newegg.ts
+++ b/src/store/model/newegg.ts
@@ -19,12 +19,12 @@ export const Newegg: Store = {
       },
     ],
     maxPrice: {
-      container: '.price-current',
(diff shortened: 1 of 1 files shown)


jef (approved):

jef:
Very good! Thank for fixing

PR #3229, opened 2025-03-11
fix: notifications to respect add to cart config

CardenB:
### Description
Twilio and pushover notifications would always use add to cart links despite config. I changed this behavior.

### Testing
Tested in prod locally 😅


MrSpookyAngel:
Line 29 for pushover also requires updating.

jef (commented):
Is the idea here that folks without auto add to cart wouldn't want the cart URL? They would want the link?

I don't think that's the right assumption, but could be swayed...

CardenB:
Yes. My motivation was that I was never successfully able to add a 5090 without the original link. 

Also, when there are load issues with the website, it’s impossible to refresh the store page from the add to cart link. You need the original link. 

It would be totally acceptable to simply have both links, as we do with the email notifications. It would be a simple change to shift to that behavior. I made the assumption that there was a single link for brevity purposes. 

At a minimum, the link behavior is not at all consistent between notification methods, so this moves toward that direction.

jef:
> At a minimum, the link behavior is not at all consistent between notification methods, so this moves toward that direction.

Ah! I do like the consistency, so thank you for that.

> It would be totally acceptable to simply have both links, as we do with the email notifications.

I could see this being the case, but I like this for the reasons you mentioned.

If you could fix up the linting issues, I'll merged this in shortly after! Thanks @CardenB 😁

jef (approved):
All fixed up. Thank you very much for your contribution. I'm sure people will love this :)

